PR #1, opened 2018-10-29
Update instructions in README.md

OleksiyRudenko:
* move example screenshot closer to where it's been described
* mention name folder when instructing to add task folder
* remove caps and final full-stops

P.S. Great tool to collect homeworks


AMashoshyna:
@OleksiyRudenko thanks!

PR #3, opened 2018-10-30
Adding KurosavaAkira student - js-dom task finished

KurosavaAkira:
Please review @yevhenorlov @OleksiyRudenko @IgorKurkov @AMashoshyna


submissions/KurosavaAkira/js-dom/index.html
```diff
@@ -0,0 +1,35 @@
...
+            <a href="javascript:void(0);" onclick="mobileMenu()"><div class="mobile-menu">menu</div></a>
+        </div>
+        <div id="main-page"></div>
+    </div>
+    
+    <script>
+        function mobileMenu() {
+            console.log('click');
+            const x = document.getElementById('menu');
+            if (x.className === '') {
```
OleksiyRudenko: Check this API https://developer.mozilla.org/en-US/docs/Web/API/Element/classList
Some collaborator may add extra classes to element `menu`, so `x.className = ''` will reset classes, which might be something collaborators wouldn't expect.

submissions/KurosavaAkira/js-dom/index.html
```diff
@@ -0,0 +1,35 @@
...
+            <a href="#kottans">Kottans - FE 2019</a>
+            <a href="#contents">Contents</a>
+            <a href="#faq">Faq</a>
+            <a href="#about">About</a>
+            <a href="javascript:void(0);" onclick="mobileMenu()"><div class="mobile-menu">menu</div></a>
+        </div>
+        <div id="main-page"></div>
+    </div>
+    
+    <script>
```
OleksiyRudenko: Why it is important to have this code within `.html`?
Are there really good reasons to not to put it in some `.js`?

submissions/KurosavaAkira/js-dom/index.js
```diff
@@ -0,0 +1,75 @@
+window.onload = () => {
```
OleksiyRudenko: This file contains some initializing code, data (model), routing, and controller.

Please, consider splitting this file into smaller chunks per code responsibility.

submissions/KurosavaAkira/js-dom/index.html
KurosavaAkira: You're right I should move it into .js
The reason why this code is present in .html is because I quickly copypasted it from w3schools.com

submissions/KurosavaAkira/js-dom/index.html
KurosavaAkira: Good point! I never thought about this before, thanks.

KurosavaAkira:
@OleksiyRudenko  Just sent a new commit with fixes. I'm not sure I did second commit in the correct way, maybe I should close this pull request and after then send new?

submissions/KurosavaAkira/js-dom/index.js
```diff
@@ -22,12 +23,10 @@ let pages = {
     },
 }
 
-let menu = () => {
+let controller = () => {
```
OleksiyRudenko: @KurosavaAkira  Wait-wait-wait! I didn't mean you have to rename anything. Probably I wan't clear.
I meant that you have code for various purposes: data containers, initialization, main code - all in a single file.
Just distribute those logically different parts across different files. E.g. `page-data.js`, `init.js` etc. Just examples, not suggestions for particular file names.
You may include those in proper order in your `index.html`.
You don't need any massive refactoring at this stage.

OleksiyRudenko: You may wonder why to split so small project across multiple files, some of which will contain like 10-20 lines.

For the sake of code manageability and readability. Code is written for other coders first of all.

Your tiny project may grow big and splitting code at that later times may become painful. So start splitting code base into modules as soon as only possible.

Do not worry about browsers that have to load multiple files, at this stage. There are tools that will optimize code for you before publishing. And you will learn about them.

KurosavaAkira: @OleksiyRudenko Ohh I see now the main idea, A few weeks ago I tried webpack exactly for this reason. Thanks a lot for the detailed review!! 

I made already splitting of js and send new commit with changes.

OleksiyRudenko:
@KurosavaAkira Well done!

Now it's easier to refer to parts of your code.

My last question regards to `page-data.js`. Why did you decide to put HTML into JS? This is definitely a legal approach, but not very SEO friendly.

Did you consider other approaches? E.g. (a) hiding/showing components by applying relevant CSS style/class or (b) employing https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template ?

P.S. You don't need to refactor anything. The above is an open question :)

KurosavaAkira:
@OleksiyRudenko Actually, I didn't think about SEO when I started to do the task, but with SEO friendly approach this task definitely becomes more exciting :) 

I think that I decided to put HTML into JS because it was the easiest and universal approach to expand the site with new HTML objects in JS. 

Thanks for the link to <template>, I never heard about them before :)

OleksiyRudenko:
Keep in mind that HTML as text in JS is probably the least maintainable than any other method. Imagine you want to involve a web designer to expand or redesign the includes you added as a text.

Summon yet another mentor to make a code review. It is always good to have at least two reviews per PR.

Approving with reservations. Give yourself a promise that you will think about code SOLIDness, scalability, readability, and maintenability.

KurosavaAkira:
@OleksiyRudenko Thanks a lot! I'll keep in mind to think about SOLIDness and other important principles. 
BTW, is this like about SOLIDness is in tasks? If not maybe it should be there, really useful article :)

@AMashoshyna Summoning one more mentor to make a code review or merge my pull requests.

OleksiyRudenko:
@KurosavaAkira , the latest commit obviously doesn't belong to task this PR is created for.
We need to revert to the state before this commit.
One of the methods is `git revert`.
Will you try it?

KurosavaAkira:
@OleksiyRudenko Tiny JS commit was reverted with my new commit

OleksiyRudenko:
@KurosavaAkira :fire:

submissions/KurosavaAkira/js-dom/init.js
```diff
@@ -0,0 +1,14 @@
+window.onload = () => {
+    router();
+    controller();
+}
+
+let mobileMenu = () => {
+    console.log('click');
+    const x = document.getElementById('menu');
+    if (x.className === '') {
```
zonzujiro: Can be replaced with `classList.toggle`

zonzujiro:
LGTM : 🔥

OleksiyRudenko:
@KurosavaAkira Grats! All done!

KurosavaAkira:
@OleksiyRudenko  Wow this is so cool!! 😃

Do you need a link to my kottans-frontend repo with a folder for badges? I can't fully understand what document do you mean.

KurosavaAkira:
@OleksiyRudenko a link to repo https://github.com/KurosavaAkira/kottans-frontend

PR #9, opened 2018-11-04
Object-Oriented JavaScript - Frogger

KurosavaAkira:
### frontend-nanodegree-arcade-game

[Demo](https://kurosavaakira.github.io/kottans-frontend/task_js_oop/frogger) |
[Code base](https://github.com/KurosavaAkira/kottans-frontend/blob/master/task_js_oop/frogger/js/app.js)

Please, review.


KurosavaAkira:
@OleksiyRudenko To solve this problem do I need to I close this PR. make `git pull upstream master` and then commit with new PR?

OleksiyRudenko:
@KurosavaAkira i am afraid this won't help. The history will stay different.

One of the methods is to identify commits sequence that is exact in both this repo and your fork and branch from the latest common commit. Then pull from upstream. Thus you will have a line if history in sync with this repo master. The downside is that this would be acceptable for your all future branching.

The hard proof method is to kill the fork, fork again, and then stick to feature branching.

If you decide following the recommendation above you will then need to store additions not merged into this repo somewhere else until new fork available

OleksiyRudenko:
This PR is closed since it's been replaced with #10.

PR #11, opened 2018-11-08
js-dom-task by feroxes

feroxes:
### js-dom-task

[Demo](https://feroxes.github.io/DOM_practical_task/) |
[Code base](https://github.com/feroxes/DOM_practical_task)

Please, review.


submissions/feroxes/js-dom/js/script.js
```diff
@@ -0,0 +1,75 @@
+function SuperHero(name, disc, img) {
+    this.name = name;
+    this.disc = disc;
+    this.img = img;
+}
+
+let ironMan = new SuperHero('Iron Man', 'Iron Man is a fictional superhero appearing in American comic books published by Marvel Comics. The character\n' +
```
zonzujiro: Instead of `'a' + 'b'` you can use [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

submissions/feroxes/js-dom/js/script.js
```diff
@@ -0,0 +1,75 @@
...
+    let childrenList = e.target.parentElement.children;
+    console.log(childrenList);
+
+    for(let i = 0; i < childrenList.length; i++){
+        childrenList[i].className = '';
+    }
+    e.target.className = 'checked';
+    
+    if(e.target.innerHTML === 'Iron Man'){
+        content.children[0].innerHTML = ironMan.name;
```
zonzujiro: Relying on order of children is a bad practice. In some time the designer, or client, or another developer, can add new elements and this changes will break your code.

Better use some sort of selectors and find elements with `querySelector`, `getElementById`, etc.

Pls, read about [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). You can reduce the amount of repeated code by moving your `innerHtml` logic to separate function, saying `addParametersToHero`.

Also, IMO, you should do this inside your class :)

zonzujiro:
Correct me if I wrong, but you need to add screenshots, according to [description](https://github.com/kottans/frontend/blob/master/tasks/js-dom.md) of task

submissions/feroxes/js-dom/js/script.js
feroxes: all done

submissions/feroxes/js-dom/js/script.js
```diff
@@ -0,0 +1,75 @@
...
+function changeMenu(e) {
+    let childrenList = e.target.parentElement.children;
+    console.log(childrenList);
+
+    for(let i = 0; i < childrenList.length; i++){
+        childrenList[i].className = '';
+    }
+    e.target.className = 'checked';
+    
+    if(e.target.innerHTML === 'Iron Man'){
```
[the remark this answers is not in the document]
feroxes: Done

submissions/feroxes/js-dom/js/script.js
```diff
@@ -0,0 +1,75 @@
...
+        content.children[0].innerHTML = ironMan.name;
+        content.children[1].innerHTML = ironMan.disc;
+        content.children[2].outerHTML = `<img src=`+ ironMan.img +`>`;
```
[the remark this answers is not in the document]
feroxes: Done.

submissions/feroxes/js-dom/js/script.js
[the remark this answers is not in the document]
feroxes: don't quite understand how to implement it

feroxes: used template literal

submissions/feroxes/js-dom/js/script.js
```diff
@@ -0,0 +1,71 @@
+function SuperHero(name, disc, img) {
+    this.name = name;
+    this.disc = disc;
+    this.img = img;
+    this.addParametersToHero = function(){
+        document.getElementById('hero__name').innerHTML = this.name;
+        document.getElementById('hero__discription').innerHTML = this.disc;
+        document.getElementById('hero__img').outerHTML = `<img id = 'hero__img' src=`+ this.img +`>`;
```
IgorKurkov: Please, use back ticks (template string) properly. You use it twice and add some concatenation instead of interpolation  syntax for instance:
\`${variable}\`
[specification of back ticks here](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/template_strings)

OleksiyRudenko:
@feroxes , please, title this and future PRs so that titles give better idea of what is inside.

In the most of the cases we add some files, so stressing on this fact is not really informative.

Thank you.

submissions/feroxes/js-dom/js/script.js
feroxes: Done!

feroxes:
All done!

OleksiyRudenko:
@IgorKurkov Please, review changes when you have a spare minute.

feroxes:
is it all right?

zonzujiro:
IMO, it is ok for now :)

feroxes:
Yuhuuu, thank`s))

submissions/feroxes/js-dom/js/script.js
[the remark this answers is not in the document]
AMashoshyna: Sorry for late reply, missed this comment.
What I meant was something like this:
```javascript
const ironManData = {
    name: 'Iron Man',
    legend: 'Iron Man is a fictional superhero appearing in American comic books'
};

let ironMan = new SuperHero(ironManData.name, ironManData.legend);
```

This looks more readable to me. But your solution is fine, this is just an idea.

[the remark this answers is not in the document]
zonzujiro: Also, I want to add to words of @AMashoshyna. You can write like this, also:

```javascript
let ironMan = new SuperHero(ironManData);

class SuperHero {
  constructor({ name, legend }) {
    // do something
  }
}
```

It will add some magic to your code - I will not understand on what **exact** arguments relay your constructor, but sometimes such stuff can be useful if you need to pass many parameters, which is more than 3 or pass optional parameters, or you just don't care about this caveat in some case.

PR #14, opened 2018-11-10
task-6 js-dom finished

linkqwd:
### task-6 js-dom

[Demo](https://linkqwd.github.io/cake-menu/)
[Code base](https://github.com/linkqwd/cake-menu)

Please, review.


AMashoshyna:
Hi!
Great job! You found a workaround to load JSON files and your layout looks nice on mobile browser.
What I would suggest to improve. Consider your application as combination of data and representation. Data in your case is texts (dessert titles and descriptions). Everything else is representation (how the data is displayed and how it behaves, for example, reacts to user clicks). Those two things (data and presentation) may change independently. For example, the client decides on redesign or adds new items to desserts list.
What you don’t want to do is to mix data and presentation. I your case this means including html tags and css classes into your json with data. Normally you would receive data from server and in case if application design changes you would need to change both client and server side.
In your data you have two arrays (titles and contents) and rely on the order of item. This is rather unusual. A more usual approach would be to get an array of items containing both title and content each. Try to rearrange as follows:

```
{
    "data": [
        {
            "title": "Carrot cake sweet roll",
            "description": "Carrot cake sesame snaps sweet roll. Chocolate bar sugar plum dessert tart",
        },
        {
            "title": "Powder oat cake marzipan brownie cake lollipop",
            "description": "…..."
        }
    ]
}
```

Also, please see my comments in code.

submissions/linkqwd/task-6_js-dom/js/main.js
```diff
@@ -0,0 +1,64 @@
...
+/////// handling events
+
+var listItems = document.querySelectorAll('.action-menu__list-item'),
+    contentItems = document.querySelectorAll('.action-menu__content-item');
+    activeListItem = '.action-menu__list-item_state_active';
+    activeContentItem = '.action-menu__content-item_state_active';
+
+listItems.forEach(function (el, index) {
+    el.addEventListener('click', function () {
+        deactivePrevItem(activeListItem);
```
AMashoshyna: Written this way you create a new function on every array iteration. Please out this function i a variable and pass this variable to `addEventListener`.

submissions/linkqwd/task-6_js-dom/js/main.js
```diff
@@ -0,0 +1,64 @@
...
+        deactivePrevItem(activeContentItem);
+        activateItem(this, index);
+    });
+});
+
+function deactivePrevItem(item) {
+    var itemToRemove = document.querySelector(item);
+    itemToRemove.classList.remove(item.slice(1));
```
IgorKurkov: about .slice(1) when you try to slice a dot - already described in previous comment

submissions/linkqwd/task-6_js-dom/js/main.js
```diff
@@ -0,0 +1,64 @@
...
+    }();
+}();
+
+var headingsHolder = document.querySelector('.action-menu__list'),
+    contentHolder = document.querySelector('.action-menu__content');
+
+for (var key in data) {
+    createFragments(key, data[key]);
+}
+
```
IgorKurkov: here first you can declare some function which will build `document.element` with `TAG, class, src`, or some else data received from your json, then build an array of elements and after that make your adjacent injection

linkqwd:
Greetings!
Thanks for reviewing my code. Your help is strongly appreciated :)
I rewrote most parts of the code considering your feedback. Hope I understood you right and did better job this time.
So in this attempt my general goals were to rebuild JSON file and make parts of code more independent one from another as well as data and presantation parts of this app.
I struggled alot with organizing JSON, my first attempt (before review) was looking something like this: http://joxi.ru/brRG57WFJbWQqm so then I decided to keep things as simple as possible, at least for start, and put content just in a string.

submissions/linkqwd/task-6_js-dom/js/main.js
linkqwd: Unfortunately i couldn't find solution to this problem;
Since I want to pass arguments in function which I use for addEventListener, i have to use anonymous function (or callback, not sure which is right) and then invoke function with desired agruments.
I guess, I could do the same using closures, but I guess it wouldn't help with this issue, coz a new lexical environment with it's own variables would be created for each function invoke?

Also, I could use event delegation to avoid that problem, but that would make my code much  bigger and complicated, at least with my current skill of coding.

linkqwd:
> ... So when I see a variable named with a verb I expect that the variable contains a function, but it doesn't, it contains function invocation result. Please check your code for such cases. You can put function in a variable, then call it and put the result into another variable, for example.

Done

AMashoshyna:
Oh, the last important thing. Take a look a function `setupSelectors`. Every time this function is called you code performs an expensive operation of querying the DOM. This function is part of click handler and will be called many times. However your data is static and the DOM elements will therefore not going to change, so calling `setupSelectors` makes no sense. I suggest you put DOM elements into `selectors` variable directly and access it from your handler.

linkqwd:
Done. 
Also, tried to implement eventDelegation to handle events, so I could solve this issue:
```
Written this way you create a new function on every array iteration. Please out this function i a variable and pass this variable to addEventListener.
```
I did strange thing on 85 line, because I need to know ordinal number of clicked element so I could activate appropriate descriptionItem (my solution with layout forced me to do this);
That would be cool if you point out how it's possible to avoid loop in this case, or it's normal practice?
Thanks

AMashoshyna:
@linkqwd very inventive solution =)

You can put an id on every `li` element. Yes, I said that we don't want to keep our data in DOM, but ids are ok and generally accepted. Normally when you get an array of items from server, those items have unique ids. For this case you can either add some id to your data, or go simple and use array index as id. So, on line 78, when you find nearest `li` element to click target, you can get this `li`'s id and it will be your ordinal number.

PR #16, opened 2018-11-16
Task DOM - completed

MitchfFirstGit:
### Document Object Model

[Demo](https://mitchffirstgit.github.io/site/) |
[Code base](https://github.com/MitchfFirstGit/site)

Please, review.


yenobi:
Okay, your solution solves the initial goal of a task, but let's make it way more prettier.

About markup:
- read [this](http://html5doctor.com) and try to make a habit to use html5 tags such as [nav](http://html5doctor.com/element-index/#nav) for navigation or [main](http://html5doctor.com/element-index/#main-desc) for the main content.

About CSS:
- global styles for the whole page is a bad habit, better set to zero margins and paddings only on blocks, where you really need this.
- try not to use tags for styling, because you can accidentally add them to places where you don't want to, use classes instead
- to make styles easier to read move all the responsive styles to the bottom of CSS-file

About js:
- try not to use id-getting, better use unique, meaningful classes / data-attributes and querySelector-method
- you can pass the name of a programmer from the event if addEventListener will be on each of list items instead of the whole list (hint: data-attributes / content of list item / identification key). this will remove cycle from your code and make it more straightforward.
- try to make logic blocks inside your script. Rearrange script with firstly removing old content and only then getting and inserting the new one
- Removing and Adding elements to the DOM is a too expensive operation, you can only change the value inside DOM elements to render different info about programmers. Investigate how to do this and implement, please.

zonzujiro:
@yenobi oi! Pls, can you take a look at changes when you will have a free time?

zonzujiro:
It is ok for now

PR #18, opened 2018-11-17
pre-OOP ready to review

linkqwd:
#### A Tiny JS World

[Code base](https://github.com/linkqwd/a-tiny-JS-world/blob/gh-pages/index.js)
[Demo](https://linkqwd.github.io/a-tiny-JS-world/)


submissions/linkqwd/index.js
```diff
@@ -0,0 +1,70 @@
...
+    this.friends = props.friends;
+    this.inhabitnatInfo = function() {
+        return JSON.stringify(this)
+        .split('')
+        .map(function (el) {
+            return el === '"' || el === '[' || el === ']' ? '' : el === ',' ? '; ' : el
+        })
+        .join('')
+        .slice(1, -1);
+    }
```
yevhenorlov: Ok, so this part is actually interesting.

The way you've implemented `this.inhabitantInfo` has the right to be, but has it's pros and cons.

Pros: if you add any other data to the constructor (for example, `this.age`), you don't need to change anything inside `this.inhabitantInfo`.

Cons: if later you need to add some data that should not be logged to `this.inhabitantInfo`, this implementation doesn't work, as it always logs every property. Also, this implementation is not very readable - as someone who looks at `this.inhabitnatInfo` for the first time, I am confused about what's going on. 

Conceptually, we are:

1. converting the separate parts of the Inhabitant into a single string (with extra symbols like brackets)

1. trying to make sense of it by iterating through and removing all the garbage.

It's probably easier to reason about parts of the `Inhabitant` when its parts are kept separate. Consider the following implementation:

```
this.inhabitantInfo = function() {
  return this.species + ';'
    + this.name + ';'
    + this.sex + ';'
    + this.arms + ';'
    + this.legs + ';'
    + this.sound + ';'
    + this.friends
}

It is verbose, as it repeats everything we have just written above, but now it is much more clear to see what we are actually showing by calling `inhabitantInfo`. If later we need to hide any field from output, we simply remove it from `inhabitantInfo`.

This is the type of decision programmers face very often - should we optimize for performance, or for readability? The choice is not always obvious, each decision has its tradeoffs.

Whether to change your implementation of `inhabitantInfo` or not is completely up to you, just make sure you consider other ways to do this.

**Comment by linkqwd** [comment]
Thanks for reviewing. I understood your point and will try to write less confusing and more readable code.
It just I wanted to meet this requirement:
```
If you know how to improve the code sample above e.g. employing #Array.join or a function that takes an object as an argument and returns a string to feed to print() then go ahead.
```
But with a lack of skill I've made the best I could :D

yevhenorlov:
Well done!
I recognised your attempt to improve the method, and it definitely gets the job done - that's why I didn't insist on you fixing it. 

There's often more than one right way to do stuff in programming, I just wanted to point out that there's almost always tradeoffs.
The point of this task is to show exactly that - and provoke thinking on the problem. You did great! 👍

linkqwd:
Thanks for inspiring words :)
Probably in my previous comment I didn't fully understand your point and I will keep in mind about tradeoffs.
It now seems even more important, as I started notice that I often confuse even myself with my code after couple days have passed.

PR #19, opened 2018-11-18
Maksym Kalashnyk added his DOM homework

kalash14:
### js-dom

[Demo](https://kalash14.github.io/dom-homework/) |
[Code base](https://github.com/kalash14/frontend-2019-homeworks/tree/dom-api-task/submissions/kalash14/js-dom)

Please, review.


submissions/kalash14/js-dom/script.js
```diff
@@ -0,0 +1,150 @@
...
+
+        let menuItem = document.createElement('li');
+        menuItem.classList.add('menu-content__item');
+        let linkMenuItem = document.createElement('a');
+        linkMenuItem.classList.add('menu-content__link');
+        linkMenuItem.setAttribute("href", "");
+        linkMenuItem.setAttribute("role", "button");
+        linkMenuItem.textContent = item.catName;
+        /* setting cat object to node in order to have access to necessary data */
+        linkMenuItem.itemData = item;
```
AMashoshyna: Here you store your data in DOM, but this approach is often seen a a bad practice. Reasons for that are, for example, the following:
- it is better and more flexible to keep your data and presentation layers independent. 
- you don't want to add random properties to DOM elements, because you don't know all possible consequences (for example, such property may already exist).

Common approach to this problem is to use some unique id. You add ids to cats and to corresponding DOM elements and retrieve data from catInformation by this id.

submissions/kalash14/js-dom/script.js
```diff
@@ -0,0 +1,150 @@
...
+    const catImage = getDOMElement().catImage;
+    const catTextInfo = getDOMElement().catTextInfo;
+
+    if (event.target.classList.contains('menu-content__link')) {
+
+        event.preventDefault();
+        let currentTarget = event.target;
+        /* getting data from current target node object 'itemData' and changing content of necessary node */
+        let currentTargetData = currentTarget.itemData;
+        catTitle.textContent = currentTargetData.catName;
```
AMashoshyna: Let's make the logic of rendering a cat reusable. You make same thing here and also when you create side menu and render default cat. I suggest you put it in a function to reduce code repetition.

AMashoshyna:
Hi @kalash14! You did a great job! Please see my comments and let me know if you need any help or explanations.

submissions/kalash14/js-dom/script.js
```diff
@@ -0,0 +1,150 @@
+let getDOMElement = () => {
+
+    /* accesing to classes of elements */
```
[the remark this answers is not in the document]
kalash14: Thanks! It was an unsuccessful experiment )))

submissions/kalash14/js-dom/script.js
```diff
@@ -0,0 +1,150 @@
...
+        mainContent: 'main-content',
+        menuWrap: 'menu-content',
+        catTitle: 'article-content__title',
+        catTexInfo: 'article-content__text',
+        catImage: 'article-content__image'
+    };
+
+    return {
+        menuIcon: document.getElementsByClassName(DOMStrings.menuIcon)[0],
+        sidebar: document.getElementsByClassName(DOMStrings.sidebar)[0],
```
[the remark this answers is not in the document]
kalash14: Ok! But I read some articles where there were compared selectors performance (getElementById, querySelector and getElementsByClassName). And querySelector was the slowest... So I decided to use getElementsByClassName.

submissions/kalash14/js-dom/script.js
```diff
@@ -0,0 +1,150 @@
...
+        menuWrap: document.getElementsByClassName(DOMStrings.menuWrap)[0],
+        catTitle: document.getElementsByClassName(DOMStrings.catTitle)[0],
+        catTextInfo: document.getElementsByClassName(DOMStrings.catTexInfo)[0],
+        catImage: document.getElementsByClassName(DOMStrings.catImage)[0]
+    }
+
+};
+
+let catsInformation = () => {
+
```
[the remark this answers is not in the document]
kalash14: Ok!

submissions/kalash14/js-dom/script.js
```diff
@@ -0,0 +1,150 @@
...
+    ];
+
+    return cats;
+
+};
+
+let renderPage = () => {
+
+    const catTitle = getDOMElement().catTitle;
+    const catImage = getDOMElement().catImage;
```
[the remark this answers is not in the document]
kalash14: Ok, nice!

submissions/kalash14/js-dom/script.js
kalash14: Ok!

submissions/kalash14/js-dom/script.js
```diff
@@ -94,16 +91,15 @@ let renderPage = () => {
...
 
-            for (let i = 0, len = cats.length; i < len; i++) {
+            /* finding match of cat with the same id as in clicked link */
 
-                let catItem = cats[i];
-                let catItemDataId = catItem.id;
-                if (currentTargetDataId === catItemDataId) {
-                    assignData(catItem);
-                }
+            cats.find((cat) => {
```
AMashoshyna: Almost, but not quite =)
Take a closer look at Array.prototype.find signature on MDN. This method accepts a function, which should return a boolean value (true if item matched the search criteria, false otherwise). Array.prototype.find returns the first element for which callback function returned true, or `undefined` if no matching element was found. In your version this callback function always return undefined, so you use it as `.forEach` basically.
You need to do something like this:
```
let cat = cats.find(cat => /* some condition here: is this the right cat? */)
if(cat) {
    renderCat(cat);
}
```

submissions/kalash14/js-dom/script.js
```diff
@@ -93,13 +93,17 @@ let renderPage = () => {
 
             /* finding match of cat with the same id as in clicked link */
 
-            cats.find((cat) => {
-
+            let currentCat = cats.find((cat) => {
                 if (cat.id === currentTargetDataId) {
-                    return renderCat(cat);
+					return true;
```
AMashoshyna: last thing! I promise =) 
What does expression `cat.id === currentTargetDataId` return?
It returns a boolean value (true/false). So just return it!

```
let currentCat = cats.find(cat => cat.id === curentTargetDataId);
```

AMashoshyna:
Great job and thanks for all the corrections! I will merge the PR, but take a look at my comment to line 97.

submissions/kalash14/js-dom/script.js
kalash14: Ok! You are absolutely right! Thanks :)

PR #20, opened 2018-11-18
DOM api task

IrynaY:
###  DOM API

[Demo](https://irynay.github.io/test/) |
[Code base](https://github.com/IrynaY/test)

Please, review.


submissions/IrynaY/js-dom/index.js
```diff
@@ -0,0 +1,23 @@
...
+  const pageContent = document.getElementById("page_content")
+  const menuContainer = document.getElementById("menu_container")
+  const menuButton = document.getElementById("menu_btn")
+
+  pageContent.innerHTML = `<img src="${content.cat.img}" class="pic">` + content.cat.content
+
+  Object.keys(content).forEach(el => {
+    const li = document.createElement("li")
+    li.innerHTML = `${content[el].title}`
+    li.addEventListener("click", function () {
```
zonzujiro: Imagine, that you have a table with a lot of elements and you need to add the listener to each cell. Something like google sheets. In such case adding a listener to each element is a bad idea, because you will create a function for every cell. Instead, you can use `event delegation` and add one listener to all table. Read about it and use this practice here.

submissions/IrynaY/js-dom/index.js
```diff
@@ -0,0 +1,23 @@
...
+      pageContent.innerHTML = `<img src="${content[el].img}" class="pic">` + content[el].content
+    })
+    menuList.appendChild(li)
+  })
+
+  menuButton.addEventListener("click", function () {
+    menuContainer.style.display !== "flex" ?  
+      menuContainer.style.display = "flex" : menuContainer.style.display = "none"
```
zonzujiro: You should use classes and `classList` API for such changes. 

This is because of several reasons:
1. Classes are more declarative. In such way, you will show "what" going on instead of "how". Such things will make code cleaner and more descriptive.
2. On every direct change of style, the browser will call `repaint` and/or `reflow` processes. Long story short - this is operations which must be made for every change of the page, so better give to browser opportunity to apply such changes in on step, just by getting all the styles from CSS.
3. Styles are matter of CSS, not JS :)

submissions/IrynaY/js-dom/style.css
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  #content{
+    width: 100%;
+  }
+}
+
+@media screen and (min-width: 800px) {
+  #menu_container{
+    display: flex !important;
```
zonzujiro: Usually, try to avoid `!important`. Why do you need it here?

submissions/IrynaY/js-dom/content.js
```diff
@@ -0,0 +1,54 @@
+const content = {
```
mgdskr: Here we have a list of similar menu items - please use appropriate data structure for them - an array. Also you should use meaningful naming like `menuItems` but not general `content` which says nothing about itself.

submissions/IrynaY/js-dom/index.js
```diff
@@ -0,0 +1,23 @@
+window.onload = function() {
+  const menuList = document.getElementById("menu");
+  const pageContent = document.getElementById("page_content")
+  const menuContainer = document.getElementById("menu_container")
+  const menuButton = document.getElementById("menu_btn")
+
+  pageContent.innerHTML = `<img src="${content.cat.img}" class="pic">` + content.cat.content
+
+  Object.keys(content).forEach(el => {
```
mgdskr: Using `Object.keys` you can't be sure about keys order and this is one more reason to use an array instead of object to keep menu items.

submissions/IrynaY/js-dom/index.js
```diff
@@ -0,0 +1,23 @@
...
+    const li = document.createElement("li")
+    li.innerHTML = `${content[el].title}`
+    li.addEventListener("click", function () {
+      pageContent.innerHTML = `<img src="${content[el].img}" class="pic">` + content[el].content
+    })
+    menuList.appendChild(li)
```
mgdskr: It is better to create a whole fragment and insert it once to the DOM than append each item separately.

submissions/IrynaY/js-dom/index.js
IrynaY: Thanks)

submissions/IrynaY/js-dom/style.css
IrynaY: interrupted another selector, i'll remove !important)

submissions/IrynaY/js-dom/index.js
IrynaY: thanks)

submissions/IrynaY/js-dom/content.js
IrynaY: thanks, fixed)

submissions/IrynaY/js-dom/index.js
IrynaY: thanks)

submissions/IrynaY/js-dom/index.js
IrynaY: try to fix))thanks)

submissions/IrynaY/js-dom/index.js
```diff
@@ -0,0 +1,26 @@
...
+  menuContainer.appendChild(ul)
+
+  ul.addEventListener("click", (evt) => {
+    if (evt.target.tagName != "LI") return
+    let element = сharacters.filter(el => {return el.id === evt.target.textContent})
+    pageContent.innerHTML = `<img src="${element[0].img}" class="pic">` + element[0].content
+  })
+
+  menuButton.addEventListener("click", () => {
+    ul.className === "visible" ? ul.className = "disable" : ul.className = "visible"
```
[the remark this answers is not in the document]
IrynaY: spasibo)

submissions/IrynaY/js-dom/index.js
```diff
@@ -0,0 +1,26 @@
...
+  сharacters.forEach(el => {
+    const li = document.createElement("li")
+    li.innerHTML = `${el.id}`
+    ul.appendChild(li)
+  })
+  menuContainer.appendChild(ul)
+
+  ul.addEventListener("click", (evt) => {
+    if (evt.target.tagName != "LI") return
+    let element = сharacters.filter(el => {return el.id === evt.target.textContent})
```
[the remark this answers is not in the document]
IrynaY: exactly. it's a arrow function. thanks)

mgdskr: If you are iterating over array of `characters` it would be more meaningful to name element `character` like `characters.filter(character => ...` Also `Array.prototype.filter` returns new array and it should be reflected in the variable name you are assigning the result to `const activeCharacters = characters.filter(character => ...)`. But in this case I would suggest to look at `Array.prototype.find` which return an element, but not array and it would be more clear `const activeCharacter = characters.find(character => ...)`

IrynaY: Understood) fixed) thanks)

IrynaY:
Guys, what should I do next in this pull request?

OleksiyRudenko:
@IrynaY please, proceed to the next task. You are not required to suspend while any PR is under revision. Apologies for the delay.

@zonzujiro @mgdskr would you please review at your earliest convenience?

PR #22, opened 2018-11-21
A tiny JS world task

IrynaY:
### A Tiny JS World

[Demo](https://irynay.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/IrynaY/a-tiny-JS-world/blob/gh-pages/index.js)

Please, review.

### DIFF
--- a/submissions/IrynaY/a_tiny_js_world/index.js
+++ b/submissions/IrynaY/a_tiny_js_world/index.js
@@ -1,9 +1,9 @@
 class Inhabitant{
-    constructor(species, name, gender, legs, speak, friends){
+    constructor(species, name, gender, speak, friends = [], legs = 2){
         this.species = species
         this.name = name
         this.gender = gender
-        this.legs = legs || 0
+        this.legs = legs
         this.speak = speak
         this.friends = friends
     }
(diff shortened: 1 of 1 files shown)


submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -0,0 +1,34 @@
+class Inhabitant{
+    constructor(species, name, gender, legs, speak, friends){
+        this.species = species
+        this.name = name
+        this.gender = gender
+        this.legs = legs || 0
```
OleksiyRudenko: Right side part of `||` expression will work if left side part is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). So, in your particular case if parameter `legs` equals to zero then `this.legs` will be zero anyway since `0 || 0 == 0`. Normally `||` expression has anything non-zero and truthy on the right side.

Please, think about defaults for Inhabitant constructor parameters. It is also sensible that params that can have default values (e.g. legs count) come after params that are unlikely to have any default values (like e.g. name). Order of function/method params not necessarily the same as the order of setting object properties.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -0,0 +1,34 @@
+class Inhabitant{
+    constructor(species, name, gender, legs, speak, friends){
```
OleksiyRudenko: See note on default values below.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -0,0 +1,34 @@
...
+        this.speak = speak
+        this.friends = friends
+    }
+    getInfo(){
+        return Object.values(this).join("; ")
+    }
+}
+
+class Human extends Inhabitant{
+    constructor(name, gender, legs, hands, speak, friends){
```
OleksiyRudenko: See note on defaults above.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -0,0 +1,34 @@
...
+        super("human", name, gender, legs, speak, friends)
+        this.hands = hands || 0
```
OleksiyRudenko: See note on `||` expressions above. I would suggest humans by default have 2 hands as we are building where all are born healthy and fully compliant to "tech specs" as provided for by mother nature.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+    getInfo(){
```
OleksiyRudenko: One of benefits of inheritance is that all class methods are inherited and there is no need to re-define those unless the implementation is different. Seems like your implementation is the same for both classes.

submissions/IrynaY/a_tiny_js_world/index.js
IrynaY: thanks

PR #28, opened 2018-11-24
Do the simple responsive menu (dom js task)

Karasb888:

### task name
https://github.com/Karasb888/docs/tree/master/docs (code files)
https://karasb888.github.io/docs/ (github page)

Please, review.


AMashoshyna:
Hi @Karasb888! You did a lot, but we need to do some more work here. 

The point of this task was to manipulate the DOM with JS. Your data (text pieces) is now hardcoded into html, which is very inflexible. To add new items or change existing ones you need to change your html. A more preferable way is to put data as in a variable (an array probably) and build your html from this data using DOM API methods like `document.createElement` and `element.appendChild`. So your initial `index.html` must be (almost) empty. 

Also please use let/const instead of var.

And .DS_Store file should not be added to version control. Use .gitignore file to prevent it.

Let me know if you need any help!

Karasb888:
Fix all bugs, replace most of var, html now generate dynamically=) Thx for review.

submissions/Karasb888/js-dom/index.js
```diff
@@ -1,80 +1,93 @@
...
+  var mobileActiveItem = 0;
+  //логика меню на мобильных девайсах
   if(document.documentElement.clientWidth <= 768){
 
-      var leftArrow = document.createElement('div');
+      let leftArrow = document.createElement('div');
       menu.appendChild(leftArrow);
       leftArrow.className = 'leftArrow';
-      leftArrow.style.display='none';
+      leftArrow.classList.add('hidden');
```
OleksiyRudenko: Based on this code fragment (38-41) and similar one below:

Please, keep in mind,
1. normally it is better to append an element to a DOM node after the element is completely "built". As every change of a mounted element causes browser to repaint at least, which is a costly operation.
2. You may add more than a single class name with a single call of 
[`#Element.classList.add()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList).

submissions/Karasb888/js-dom/index.js
```diff
@@ -1,80 +1,93 @@
...
+          menuItems[mobileActiveItem - 1].classList.remove('hidden');
+          contentItems[mobileActiveItem].classList.add('hidden');
+          contentItems[mobileActiveItem - 1].classList.remove('hidden');
         }
 
-        activeMobile--;
+        mobileActiveItem--;
+        //на первом слайде убираю стрелку влево, на последнем стрелку вправо
+        mobileActiveItem === 0 ? leftArrow.classList.add('hidden') : leftArrow.classList.remove('hidden');
+        mobileActiveItem  === menuItems.length - 1 ? rightArrow.classList.add('hidden') : rightArrow.classList.remove('hidden');
```
OleksiyRudenko: This is legal use of ternary, true. However, normally ternary is used when its returned result is used to assign to a variable or as an argument of a different operation or function.

Last time there was an assignment in this code fragment, and now there is no any.

You do not need to fix anything here.

Just think over the equivalent (illustrating the possible options of ternaries use):
`rightArrow.classList[mobileActiveItem ? 'remove' : 'add' ]('hidden');`

submissions/Karasb888/js-dom/index.js
```diff
@@ -1,80 +1,93 @@
...
+        menuItems[mobileActiveItem + 1].classList.remove('hidden');
+        contentItems[mobileActiveItem].classList.add('hidden');
+        contentItems[mobileActiveItem + 1].classList.remove('hidden');
       }
 
-      activeMobile++;
+      mobileActiveItem++;
+      //на первом слайде убираю стрелку влево, на последнем стрелку вправо
+      mobileActiveItem  === menuItems.length - 1 ? rightArrow.classList.add('hidden') : rightArrow.classList.remove('hidden') ;
+      mobileActiveItem  === 0 ? leftArrow.classList.add('hidden') : leftArrow.classList.remove('hidden') ;
```
OleksiyRudenko: There must be some way to optimize a number of code blocks above. A bit too much of repetition.

You are not required (by me in this task at least) to fix this.

Just think over it when you have time for this.

PR #36, opened 2018-11-27
Memory – Pair Game

Oodmincheg:

### Memory – Pair Game

[Code base](https://github.com/Oodmincheg/troll-memory-game) | [Demo](https://oodmincheg.github.io/troll-memory-game/)

Please, review.

### DIFF
--- a/submissions/Oodmincheg/memory-pair-game/index.js
+++ b/submissions/Oodmincheg/memory-pair-game/index.js
@@ -1,0 +1,126 @@
+"use strict";
+//array of our cards
+let list = [
+  "./img/sad.png",
+  "./img/dad.png",
+  "./img/indiffirent.png",
+  "./img/pfff.png",
+  "./img/troll.png",
+  "./img/rage.png",
+  "./img/yellow.png",
+  "./img/scream.png"
+];
+//double our cards in the array
+list = list.concat(list);
+
+const GAME_BOARD = document.getElementById("content");
+const WINNER_STRING = "OK, dude. You are the winner. ";
+const TIME_OF_BACK_FLIP = 1200;
+const TIME_OF_HIDDING = 800;
+let keepPlaying, secondClick, nextMoveFlag, previousFlipper, previousImage;
+
+//funtcion to flip one or more fliper containers
+const flip = function() {
+  [...arguments].forEach(flipper => {
+    flipper.classList.toggle("flip");
+  });
+};
+
+//function that starts next move and handle game over situation
+const nextMove = () => {
+  if (keepPlaying) {
+    secondClick = false;
+    previousFlipper = null;
+    nextMoveFlag = true;
+  } else {
+    GAME_BOARD.textContent = WINNER_STRING;
+    let newGameButton = document.createElement("button");
+    newGameButton.textContent = "New Game";
+    newGameButton.addEventListener("click", newGame);
+    GAME_BOARD.appendChild(newGameButton);
+  }
+};
+
+// hide matches from page, and count how many cards are left in the game
+function removeMatches(currentFlipper) {
+  currentFlipper.classList.add("hidden");
+  previousFlipper.classList.add("hidden");
+  keepPlaying -= 2;
+  nextMove();
+}
+
+//function that handle moves(clicks on the cards)
+const move = event => {
+  const currentFlipper = event.currentTarget;
+  // prettier-ignore
+  const currentImage = event.currentTarget.querySelector("img").getAttribute("src");
+
+  //check the click on the same card, click on not hidden element and possibility of opening cards
+  if (
+    currentFlipper != previousFlipper &&
+    !currentFlipper.classList.contains("hidden") &&
+    nextMoveFlag
+  ) {
+    flip(currentFlipper);
+    if (secondClick) {
+      nextMoveFlag = false;
+      if (currentImage != previousImage) {
+        setTimeout(() => {
+          flip(previousFlipper, currentFlipper);
+          nextMove();
+        }, TIME_OF_BACK_FLIP);
+      } else {
+        setTimeout(() => {
+          removeMatches(currentFlipper);
+        }, TIME_OF_HIDDING);
+      }
+    } else {
+      previousFlipper = currentFlipper;
+      previousImage = currentImage;
+      secondClick = true;
+    }
+  }
+};
+
+const newGame = () => {
+  //shuffle array of cards
+  list.sort(function() {
+    return 0.5 - Math.random();
+  });
+
+  //set game's variables to the start values
+  GAME_BOARD.textContent = "";
+  keepPlaying = list.length;
+  secondClick = false;
+  nextMoveFlag = true;
+
+  //render our cards on the page(back sides)
+  list.forEach(img => {
+    //create blank side of the card
+    const blank = document.createElement("div");
+    blank.classList.add("front");
+
+    //create our flipper container
+    let flipper = document.createElement("div");
+    flipper.classList.add("flipper");
+
+    //add event listnee to our flipper container
+    flipper.addEventListener("mousedown", event => {
+      move(event);
+    });
+
+    //create image on back side of our card
+    let image = document.createElement("img");
+    image.setAttribute("src", img);
+    image.classList.add("back");
+
+    //add back and front of our card to the flipper
+    flipper.appendChild(image);
+    flipper.appendChild(blank);
+
+    //add aour flipper to the page
+    GAME_BOARD.appendChild(flipper);
+  });
+};
+
+newGame();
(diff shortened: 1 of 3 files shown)


submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+//double our cards in the array
+list = list.concat(list);
+
+const GAME_BOARD = document.getElementById("content");
+const WINNER_STRING = "OK, dude. You are the winner. ";
+const TIME_OF_BACK_FLIP = 1200;
+const TIME_OF_HIDDING = 800;
+let keepPlaying, secondClick, nextMoveFlag, previousFlipper, previousImage;
+
+//funtcion to flip one or more fliper containers
```
zonzujiro: Comments must detail **why** something happens, not **what**. Or you will just add redundant and useless comments.

Also, you have misprint there.

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+function removeMatches(currentFlipper) {
+  currentFlipper.classList.add("hidden");
+  previousFlipper.classList.add("hidden");
+  keepPlaying -= 2;
+  nextMove();
+}
+
+//function that handle moves(clicks on the cards)
+const move = event => {
+  const currentFlipper = event.currentTarget;
```
zonzujiro: IMO, is better. But this is totally up to you.

```javascript
const move = ({ currentTarget }) => 
```

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+  // prettier-ignore
+  const currentImage = event.currentTarget.querySelector("img").getAttribute("src");
```
zonzujiro: `currentFlipper.querySelector...`?

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+      previousFlipper = currentFlipper;
+      previousImage = currentImage;
+      secondClick = true;
+    }
+  }
+};
+
+const newGame = () => {
+  //shuffle array of cards
+  list.sort(function() {
```
zonzujiro: If you use ES6, use it everywhere. There are cases when you will need local `this`, but this is a rare case.

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+const GAME_BOARD = document.getElementById("content");
+const WINNER_STRING = "OK, dude. You are the winner. ";
+const TIME_OF_BACK_FLIP = 1200;
+const TIME_OF_HIDDING = 800;
+let keepPlaying, secondClick, nextMoveFlag, previousFlipper, previousImage;
+
+//funtcion to flip one or more fliper containers
+const flip = function() {
+  [...arguments].forEach(flipper => {
```
xanf: Accessing arguments is bad idea in modern JavaScript. It's better to use
```js
const flip = function(...args) {
  args.forEach(...)
}
```

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+    GAME_BOARD.textContent = WINNER_STRING;
+    let newGameButton = document.createElement("button");
+    newGameButton.textContent = "New Game";
+    newGameButton.addEventListener("click", newGame);
+    GAME_BOARD.appendChild(newGameButton);
+  }
+};
+
+// hide matches from page, and count how many cards are left in the game
+function removeMatches(currentFlipper) {
```
xanf: It's not good that currentFlipper is passed like an argument to this function and previousFlipper is taken ...from elsewhere.

It's generally a good idea to keep functions pure - that means also that they are accessing only their arguments

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+    //create blank side of the card
+    const blank = document.createElement("div");
+    blank.classList.add("front");
+
+    //create our flipper container
+    let flipper = document.createElement("div");
+    flipper.classList.add("flipper");
+
+    //add event listnee to our flipper container
+    flipper.addEventListener("mousedown", event => {
```
zonzujiro: `flipper.addEventListener("mousedown", move)`

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+      move(event);
+    });
+
+    //create image on back side of our card
+    let image = document.createElement("img");
+    image.setAttribute("src", img);
+    image.classList.add("back");
+
+    //add back and front of our card to the flipper
+    flipper.appendChild(image);
```
zonzujiro: You can use [append](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append) because of multi-arity. F**k the IE!

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+  "./img/scream.png"
+];
+//double our cards in the array
+list = list.concat(list);
+
+const GAME_BOARD = document.getElementById("content");
+const WINNER_STRING = "OK, dude. You are the winner. ";
+const TIME_OF_BACK_FLIP = 1200;
+const TIME_OF_HIDDING = 800;
+let keepPlaying, secondClick, nextMoveFlag, previousFlipper, previousImage;
```
xanf: keepPlaying is not the BEST naming for variable. The reason is other dev (like me) while reviewing code is expecting it to be boolean flag, but this is actually remaining active tiles. Maybe something like `remainingCards` sounds better?

OleksiyRudenko:
@Oodmincheg didn't look at the code yet. Could you please give this PR more specific title? All PRs do normally add some code.

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+//funtcion to flip one or more fliper containers
+const flip = function() {
+  [...arguments].forEach(flipper => {
+    flipper.classList.toggle("flip");
+  });
+};
+
+//function that starts next move and handle game over situation
+const nextMove = () => {
```
zonzujiro: `performNextMove` will be better.

submissions/Oodmincheg/memory-pair-game/style.css
```diff
@@ -0,0 +1,83 @@
...
+  height: 100%;
+}
+
+h1 {
+  font-family: "Comic Sans MS", cursive, sans-serif;
+  text-align: center;
+  font-size: 20px;
+  margin: auto;
+}
+.flipper {
```
zonzujiro: You already have `.flipper` at line 14

submissions/Oodmincheg/memory-pair-game/style.css
```diff
@@ -0,0 +1,83 @@
...
+  transition: transform 0.6s;
+  transform-style: preserve-3d;
+
+  position: relative;
+}
+
+/* hide back of pane during swap */
+.front,
```
zonzujiro: Same here - line 21

OleksiyRudenko:
@Oodmincheg thanks.
Please, also change PR's title itself
image

...so in the list of PRs it also gives some idea on what's inside
image

Oodmincheg:
Fixed, but I'm not sure about few details :)

submissions/Oodmincheg/memory-pair-game/index.js
Oodmincheg: > IMO, like this is better. But this is totally up to you.
> 
> ```js
> const move = ({ currentTarget }) => 
> ```
I've understood what I meant(and actually wrote this way), but than figured out some bug, and have had to add event.preventDefault() to this function

zonzujiro: 👍

submissions/Oodmincheg/memory-pair-game/index.js
Oodmincheg: Create big function with more clear named vairables(imo clear named...)

submissions/Oodmincheg/memory-pair-game/index.js
```diff
@@ -0,0 +1,119 @@
...
+  let blankSide = document.createElement("div");
+  blankSide.classList.add("front");
+  let imageSide = document.createElement("img");
+  imageSide.setAttribute("src", img);
+  imageSide.classList.add("back");
+  card.append(imageSide, blankSide);
+  return card;
+};
+
+const newGame = () => {
```
zonzujiro: Name of functions - verbs. Name of variables - nouns. It **is does** matter, especially in the large projects, when you want to spend a less time to investigate what code doing.

But for now, it is ok.

Oodmincheg: Got it.

Oodmincheg:
> pls resolve fixed discussions

Done.

PR #42, opened 2018-11-28
Memory game

MitchfFirstGit:
###Memory – Pair Game

[Demo](https://mitchffirstgit.github.io/Memory_Game_Kottans/) |
[Code base](https://github.com/MitchfFirstGit/Memory_Game_Kottans)

Please, review.


submissions/MitchfFirstGit/Memory_Pair_Game/add.js
```diff
@@ -0,0 +1,79 @@
+const CARDS = document.querySelectorAll('.card');
+const TIME_SHOW_MATCH_CARDS = 500;
+const TIME_KOTTANS = 11;
+let firstGuessCard, secondGuessCard, startTime, endTime, gameTime, clickCard;
+let activeCard = false;
+let lockAttempt = false;
+let gameResult = 0;
+let gamePairs = 6;
+function flip(){
+  clickCard = this;
```
zonzujiro: Well... it is almost the same :) 

Usualy you must can take event target, in your case card, from an [event](https://developer.mozilla.org/en-US/docs/Web/Events/click) itself. So no need in `this`.

I think this one will be better:

```javascript
const flip = ({ target }) => {

}
```

submissions/MitchfFirstGit/Memory_Pair_Game/add.js
```diff
@@ -0,0 +1,79 @@
...
+  if(firstGuessCard.firstElementChild.src === secondGuessCard.firstElementChild.src){
+      firstGuessCard.removeEventListener('click', flip);
+      secondGuessCard.removeEventListener('click', flip);
+      gameResult++;
+      setTimeout(()=>{
+        firstGuessCard.classList.add('hidden');
+        secondGuessCard.classList.add('hidden');
+        if(gameResult === gamePairs){
+          endTime = new Date().getTime();
+          const convertToSeconds = ms => (ms / 1000).toFixed(1);
```
zonzujiro: You will recreate this function on every timeout and this is not necessary. Move variable to global scope pls.

submissions/MitchfFirstGit/Memory_Pair_Game/add.js
```diff
@@ -0,0 +1,79 @@
...
+}
+function shuffleCards(){
+  alert(`If you think you are Kottan then you have ${TIME_KOTTANS} seconds to complete this game`);
+  for(let i = 0; i<CARDS.length;i++){
+    let ramdomPos = Math.floor(Math.random() * 12);
+    CARDS[i].style.order = ramdomPos;
+    CARDS[i].lastElementChild.classList.add('hidden');
+  }
+  setTimeout(function(){
+    for(let i = 0; i < CARDS.length; i++){
```
zonzujiro: Wrong spacing. Try to use https://prettier.io/playground/

Also, prefer array methods to loops, because methods are declarative. Here you can use `Array#forEach`

PR #45, opened 2018-11-29
Building a Tiny JS Worldl

maxovsanyuk:
### Building a Tiny JS World 

[Code base](https://github.com/maxovsanyuk/kottans-frontend/blob/master/task_js-pre-oop/index.js) [demo](https://maxovsanyuk.github.io/a-tiny-JS-world/)

Please, review.

### DIFF
--- a/submissions/maxovsanyuk/a-tiny-JS-world/index.js
+++ b/submissions/maxovsanyuk/a-tiny-JS-world/index.js
@@ -15,7 +15,7 @@ class Inhabitant{
         this.name = `${name.charAt(0).toUpperCase() + name.slice(1)}`;
         this.gender = gender;
         this.legs = legs;
-        this.hands = +hands === null ? '0': hands;
+        this.hands = hands === 0 ? '0': undefined;
         this.introduce = introduce;
         this.expression = `<b>General information</b> ( ${specie}, ${gender} ). Hello my name is: <b>${this.name}</b>.  I'm a ,and I have: <b>${this.legs}</b> legs and <b>${this.hands}</b> hands. About myself: <b>${this.introduce}</b>`;
     }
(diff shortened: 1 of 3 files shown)


submissions/maxovsanyuk/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,55 @@
...
+// Define your objects here
+
+
+class Inhabitant{
+    constructor(specie, name, gender, legs, hands, introduce){
+        this.specie = specie;
+        this.name = `${name.charAt(0).toUpperCase() + name.slice(1)}`;
+        this.gender = gender;
+        this.legs = legs;
+        this.hands = +hands === null ? '0': hands;
```
OleksiyRudenko: Nothing, even `null`, will make the condition to be true. E.g. `+null === 0`.
This means that your construction can be effectively optimized to `this.hands = hands;`
Why do you need this check? Or did you mean that anything [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) should be converted into `'0'`? This would be OK. But you then need to fix the comparison.

Please, fix it. Just do not over-complicate things. The aim of this task is different, yet mistakes of this kind should have been fixed.

maxovsanyuk:
```suggestion
        this.hands = hands ===  0 ? '0':  undefined;
```

submissions/maxovsanyuk/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,55 @@
...
+// Define your objects here
+
+
+class Inhabitant{
+    constructor(specie, name, gender, legs, hands, introduce){
+        this.specie = specie;
+        this.name = `${name.charAt(0).toUpperCase() + name.slice(1)}`;
+        this.gender = gender;
+        this.legs = legs;
+        this.hands = hands === 0 ? '0': undefined;
```
OleksiyRudenko: I am afraid this won't work as it is expected to. Anything not strictly equal to numeric zero will result in `undefined`. So, humans will have `undefined` count of hands.

Anyway you do not perform any checks as to legs. Let's put things simple.

If you decide to fix, then, please, try different approach on your Tiny JS World fork repo. When happy, update demo and also update code here.

maxovsanyuk:
```suggestion
        this.hands = hands;
```

PR #51, opened 2018-12-02
js-dom - done

justdevway:
### task name

[Demo](https://justdevway.github.io/frontend-2019-homeworks/) |
[Code base](https://github.com/justdevway/kottans-frontend/tree/master/task_js_dom)

Please, review.


submissions/justdevway/js-dom/index.html
```diff
@@ -0,0 +1,56 @@
...
+                        <li class="item js-item js-item--first">
+                            <a href="#" class="item__link  js-item__link isActive" data-link="1">популярные</a>
+                        </li>
+                    </ul>
+                </nav>
+                <main class="results js-results"></main>
+            </div>
+        </div>
+    </div>
+    <div class="modal js-modal" style="display: block">
```
AMashoshyna: Writing styles in html directly in often considered bad practice. At least it is no very good to mix classes and inline styles. Please add a class and put the styles in this class instead.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,234 @@
+document.addEventListener("DOMContentLoaded", function (event) {
+    var API_KEY = '95ca7e90ff0bb4cbff4bc6b45390f12d';
```
AMashoshyna: `var` keyword is technically not deprecated, but please try to avoid it. `let` and `const` are preferable.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,234 @@
...
+    var links = {
+        genre: 'https://api.themoviedb.org/3/genre/movie/list?api_key=' + API_KEY + '&language=ru-RU',
+        popular: 'https://api.themoviedb.org/3/movie/popular?api_key=' + API_KEY + '&language=ru-RU',
+        forGenre: 'https://api.themoviedb.org/3/discover/movie?api_key=' + API_KEY + '&language=ru-Ru&sort_by=popularity.desc&include_adult=false&include_video=false&page=1&with_genres='
+    };
+
+    function appendContent(link, cb) {
+        // initModal();
+        showModal();
```
AMashoshyna: This function does two very different things: it shows modal and gets data. It should have only one responsibility.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,234 @@
...
+        xhr.send();
+
+        xhr.onreadystatechange = function () { // (3)
+            if (xhr.readyState != 4) return;
+
+            if (xhr.status != 200) {
+                console.error(xhr.status + ': ' + xhr.statusText);
+            } else {
+                console.dir(JSON.parse(xhr.responseText, 2));
+                cb(xhr.responseText);
```
AMashoshyna: Every callback function which receives `xhr.responseText`, is parsing the data. If you parse the data before calling the callback, you only need to do it once.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+    }
+
+
+    // TODO: write preloader
+
+    function buildContent(data) {
+        var data = JSON.parse(data, 2).results;
+        var container = document.getElementsByClassName('js-results')[0];
```
AMashoshyna: take a look at `document.querySelector` method.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,234 @@
...
+        container.innerHTML = '';
+
+        data.forEach(function (el, index) {
+            var item = document.createElement('article');
```
AMashoshyna: Wow, this is a very long function! Please split it in parts and try to reduce code duplication as much as possible. Make a function (or several function) which receives data and returns DOM element.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,234 @@
...
+            var link = document.createElement('a');
+            link.classList.add('item__link', 'js-item__link');
+            link.setAttribute('data-link', index);
+            link.setAttribute('data-menu-id', el.id);
+            link.append(document.createTextNode(el.name));
+            item.append(link);
+            container.append(item);
+            item.classList.add('item');
+            if (index === 8) {
+                var showMore = document.createElement('li');
```
AMashoshyna: Numbers in code without explanation are called "magic numbers". Two problems:
- it is hard to understand what 8 means and what this code does
- you use 8 on line 105 again. If you decide to change this value one day, you will have to change it in two places and may forget something.
Let's put this 8 into a properly named constant.

submissions/justdevway/js-dom/index.html
justdevway: okey, thanks

submissions/justdevway/js-dom/js/main.js
justdevway: I tried to understand  callback function more and not use ES6 in this task. 
Can I left this variables ( clean old style :) ), or remove as you said ?

submissions/justdevway/js-dom/js/main.js
justdevway: So one function that take data and other that show modal, I correct understand ?

submissions/justdevway/js-dom/js/main.js
justdevway: okay, mistake, thanks :)

submissions/justdevway/js-dom/js/main.js
justdevway: Can you say reason why better use `document.querySelector` instead of `document.getElementsByClassName` ? 
Sorry, I really don't know this information.

submissions/justdevway/js-dom/js/main.js
justdevway: so, one function that build item, one that get data and loop it, I correct understand ?

submissions/justdevway/js-dom/js/main.js
justdevway: My mistake, thanks. 
Thanks for your time :)

submissions/justdevway/js-dom/js/main.js
AMashoshyna: Sure, if you are doing it consciously, it's ok.

submissions/justdevway/js-dom/js/main.js
AMashoshyna: Nothing to feel sorry about, we are here to learn =)
`document.getElementsByClassName` returns a collection of all elements matching the selector, but you only need the first match, right? `document.querySelector` returns the first match, so it shows your intention more explicitly, is more expressive and shorter too. So, in my opinion, it fits this case better, but your choice is ok.

submissions/justdevway/js-dom/js/main.js
AMashoshyna: Yes. I think you can show modal before calling `appendContent` function.

submissions/justdevway/js-dom/js/main.js
zonzujiro: @justdevway decide for yourself which way of splitting is better. In general, you should separate logic as much as possible.

One of the examples is how you proposed.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,245 @@
...
+            poster.setAttribute('alt', el.title);
+            overview.classList.add('result__description');
+            overview.append(document.createTextNode(el.overview));
+            date.classList.add('result__data');
+            date.append(document.createTextNode('Дата релиза: ' + el.release_date));
+            date.dateTime = el.release_date
+            header.append(title);
+            item.append(header);
+            content.append(poster);
+            content.append(overview);
```
zonzujiro: `append` is a multi-arity function, so you can write like this: `content.append(poster, overview, ...)`

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,245 @@
...
+        var container = document.getElementsByClassName('js-list')[0];
+        var showLess = showLessButton();
+
+        data.forEach(function (el, index) {
+            var item = document.createElement('li');
+            var link = document.createElement('a');
+            link.classList.add('item__link', 'js-item__link');
+            link.setAttribute('data-link', index);
+            link.setAttribute('data-menu-id', el.id);
+            link.append(document.createTextNode(el.name));
```
zonzujiro: Why you have used `createTextNode` here?

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,245 @@
...
+            }
+        });
+
+        container.append(showLess);
+        initBurger();
+        initNavButton();
+        initNavLinks();
+    }
+
+    function  showMoreButton () {
```
zonzujiro: Redundant space 👀

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,245 @@
...
+        var tempLink = document.createElement('a');
+        showLess.classList.add('item');
+        tempLink.classList.add('item__link', 'item__link--showLess', 'js-item__link--showLess');
+        tempLink.append(document.createTextNode("скрыть"));
+        showLess.append(tempLink);
+        return showLess;
+    }
+
+    function changeImgs(cb) {
+        var imgs = Array.prototype.slice.call(document.getElementsByClassName('js-result__img'));
```
zonzujiro: Use `Array.from` instead :)

AFAIK, you can use `NodeList.forEach` without converting it to an array. Pls, check this information.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,245 @@
...
+            el.setAttribute('src', src);
+        });
+
+        cb();
+    }
+
+    function initBurger() {
+        var burger = document.getElementsByClassName('js-toggle__burger')[0];
+        burger.addEventListener('click', function (e) {
+            var nav = document.getElementsByClassName('js-nav')[0];
```
zonzujiro: Same with `querySelector`

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,245 @@
...
+                this.parentElement.classList.add('isHide');
+            });
+        }
+
+        if(showLessButton) {
+            showLessButton.addEventListener('click', function (e) {
+                e.preventDefault();
+                var items = document.getElementsByClassName('item');
+                Array.prototype.slice.call(items).forEach(function (el, index) {
+                    if (index === 10) {
```
zonzujiro: This is a bad pattern because:
1. What you will do for elements `[0...9]`?
2. Such checks are redundant because you do an unnecessary job for every step. 

Instead of checking, you can write something like this:

```
const [ first, ...tail ] = Array.from(items).slice(10);

first.classList.remove('isHide')
rest.forEach(el => el.classList.add('isHide'));
```

But this is anyway is a bad code, because you need to hide your elements one-by-one. Instead, you should try to add some class to a parent of this elements, and then hide needed children with CSS only.

submissions/justdevway/js-dom/js/main.js
```diff
@@ -0,0 +1,245 @@
...
+                });
+            });
+        }
+    }
+
+    function initTitleButton() {
+        var titleLinks = Array.prototype.slice.call(document.getElementsByClassName('js-result__title__link'));
+        var modalWindow = document.getElementsByClassName('js-modal')[0];
+        titleLinks.forEach(function (el, i) {
+            el.addEventListener('click', function (e) {
```
zonzujiro: You can use event delegation here

submissions/justdevway/js-dom/js/main.js
justdevway: okay, thanks :)

submissions/justdevway/js-dom/js/main.js
justdevway: I will remove it, thanks

submissions/justdevway/js-dom/js/main.js
justdevway: Sorry, don't understand question

submissions/justdevway/js-dom/js/main.js
justdevway: Okay, I will check it. Don't know about it, thanks

submissions/justdevway/js-dom/js/main.js
justdevway: Discuss about it with @AMashoshyna , decided on this project use .getElementsByClassName in other I will use querySelector

submissions/justdevway/js-dom/js/main.js
justdevway: Okay, thanks, I will read more about it

AMashoshyna:
Hi @justdevway! You did a great job! I will merge the PR. As possible improvements, consider the following:
- in `navLinks` (line 235) event delegation can be used too. Generally, when you want to add an event listener in a loop, you propable need event delegation
- using inline styles and writing to element.style directly has many disadvantages, code maintenance among them. Better to stick with css classes. Add `hidden` class and use it in js instead of writing `element.style.display = '';`

PR #54, opened 2018-12-03
Oop exercise

IrynaY:

### OOP exercise (update A Tiny JS World)

[Demo](https://irynay.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/IrynaY/a-tiny-JS-world)

Please, review.


submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -13,19 +14,38 @@ class Inhabitant{
...
+}
+
+class CatWoman extends Cat{
+    constructor(name, gender, hands = 2, lags = 2, friends){
+        super(name, gender, hands, lags, friends)
+        this.species = "Cat-woman"
+    }
+}
+
+const man = new Human("Jack", "male", 2, 2,  "AAAAaaaargh!!!", ["Tom", "Bob"])
```
OleksiyRudenko: Looks like [default params](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) do not work well under current setup.
Defaults are powerful device and used often. Please, fix to skill up.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -13,19 +14,38 @@ class Inhabitant{
...
+const woman = new Human("Lucy", "female", 2, 2,  "Holla!", ["Kittie", "Bob"])
```
OleksiyRudenko: You were courageous to take an optional challenge of defining friends. Please, note that current implementation lists some strings that are presumably names (of imaginary friends?) rather than **existing** inhabitants.
This challenge is optional but you will definitely skill up when find a solution.
It is not rare case in software development when you need to cross-refer to other objects.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -13,19 +14,38 @@ class Inhabitant{
...
+const cat = new Cat("Fluffie", "female", 4, ["Lucy"])
+const dog = new Dog("Tim", "male", 0, 4, ["Jack"])
+const cat_woman = new CatWoman("Kittie", "female", 2, 2, ["Kittie", "Fluffie"])
```
OleksiyRudenko: Do you really need have that specifically named variables?
What if your next assignment to populate the world with thousands of inhabitants?
Please, employ some better scalable approach.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -13,19 +14,44 @@ class Inhabitant{
...
-const cat = new Inhabitant("cat", "Fluffie", "female", "meeeoooOOOoooow!!!", 4)
-const catWoman = new Human("Kittie", "female", cat.speak, null , 2, 2)
 
-print(dog.getInfo())
-print(woman.getInfo())
-print(man.getInfo())
-print(cat.getInfo())
-print(catWoman.getInfo())
+class Dog extends Inhabitant{
+    constructor(name, gender, friends, hands, lags = 4){
```
OleksiyRudenko: With this params order whenever one wants to override a dog's number of legs on creation she/he also forced to define a number of hands (zero or undefined, whatever), which is confusing. Please, find a better solution.
Same refers to `Cat`.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -13,19 +14,44 @@ class Inhabitant{
...
+}
+
+class CatWoman extends Cat{
+    constructor(name, gender, friends, hands = 2, lags = 2){
+        super(name, gender, friends, hands, lags)
+        this.species = "Cat-woman"
+    }
+}
+
+const inhabitant_1 = new Human("Jack", "male", "AAAAaaaargh!!!")
```
OleksiyRudenko: Still not very scalable. And forces to print named items one by one.
Think over the following snippet:
```
let friendMakers = [];
let friendMakersCounter = 0;
const inhabs = [ friendMakers[frinedMakersCount++] = new Hum(...), new Cat(...),  friendMakers[frinedMakersCount++] = new Hum(...) ];

....
friendMakers[0].friends = [....];
...
inhabs.forEach(inhab => print(...));
```

Not perfect yet, just to give you some ideas.

If you want to have named items in `friendMakers` you might use an object literal instead of an array.

submissions/IrynaY/a_tiny_js_world/index.js
```diff
@@ -13,19 +14,44 @@ class Inhabitant{
...
+const inhabitant_2 = new Human("Lucy", "female", "Holla!")
+const inhabitant_3 = new Cat("Fluffie", "female")
+const inhabitant_4 = new Dog("Tim", "male")
+const inhabitant_5 = new CatWoman("Kittie", "female")
+
+inhabitant_1.friends = [inhabitant_2.name]
+inhabitant_2.friends = [inhabitant_3.name, inhabitant_5.name]
```
OleksiyRudenko: Are friends inhabitants or just inhabitants' names?
You may also need to fix `.getInfo()` method.

submissions/IrynaY/a_tiny_js_world/index.js
IrynaY: @OleksiyRudenko Hi! Thanks for hint)
I had a question about the friendMakers array. I don't quite understand his purpose)
I think to keep  inhabitants in `inhabs` array, without using `friendMakers`, for example
`const inhabs = [
    new Hum(...),
    new Cat(...),
    new Dig(...)
];`

Can I get a hint on what to use `friendMakers` for? )

PR #61, opened 2018-12-04
add js-dom homework

Eugene50:
###Document Object Model

[Demo](https://eugene50.github.io/Test/) |
[Code base](https://github.com/Eugene50/Test)

Please, review.


OleksiyRudenko:
@Eugene50 thank you for the submission. Can you, please, complete the initial PR comment so it provides for required information accurately?

Eugene50:
Thanks for amendment!!! i am now in Japan and now here the deep nigth)

submissions/Eugene50/js-dom/app.js
```diff
@@ -0,0 +1,33 @@
+window.onload = function () {
+  const container  = document.getElementById("container");
```
AMashoshyna: Take closer look at spaces and indentations. This may seem unsubstantial now, but ruins first impression. Here - multiple spaces before `=`

submissions/Eugene50/js-dom/app.js
```diff
@@ -0,0 +1,33 @@
...
+  var content = document.getElementById("content");
+  var menu = document.getElementById("menu");
+
+
+const x = document.createElement("h1");
+x.textContent = 'Oboe World';
+document.body.insertBefore(x, document.body.firstChild);
+
+data.forEach(function (index) {
+
```
AMashoshyna: empty line and irregular indentations in this function body

AMashoshyna: `index` is a rather misleading name. Thing is, second parameter of `forEach` is actually index of the element in the array, so I'm confused whether it's index or array element. `item` or `element` maybe?

submissions/Eugene50/js-dom/app.js
```diff
@@ -0,0 +1,33 @@
...
+  const div = document.createElement("p");
+div.innerHTML = index.name;
+container.appendChild(div);
+});
+
+container.addEventListener('click', function (el) {
+
+if(el.target.tagName != 'p') {
+ var currentOboe = data.find(index => index.name === el.target.textContent)
```
AMashoshyna: Relying on layout details (like tag name or textContent) is risky. What if design changes and you will rewrite your html? Then you would need to change js too. 
In such cases it is reasonable to provide element with an id and match by id.

Eugene50:
I have corrected my code! Please checkout)

Eugene50:
I have corrected my code! Please checkout)

AMashoshyna:
@Eugene50 Opps, something went wrong, because I can't see any new commits. Are you sure you have pushed your changes?

Eugene50:
yes, sorry, i`m pusshing just to my repo! Now add new commits)

AMashoshyna:
Yes, now it's ok!
I will merge the PR, but as possible improvement consider using let/const instead of var's.
Also, as I said, relying on element order (as in line 8) is risky, because you may need to change layout structure one day. Same concerns tag names (line 18). It is better to put some explicit identifiers of elements (like ids or classes). But otherwise great job!

PR #80, opened 2018-12-08
Add Frogger Game

denislukianenko:
### Frogger

[Demo](https://denislukianenko.github.io/frogger/) |
[Code base](https://github.com/denislukianenko/frogger)

Please, review.


submissions/denislukianenko/js-oop-frogger/js/app-ref.js
```diff
@@ -0,0 +1,125 @@
...
+// Update the enemy's position, required method for game
+// Parameter: dt, a time delta between ticks
+Enemy.prototype.update = function (dt) {
+
+    // Multiplies the speed by the dt parameter on the x axis
+    this.x += this.speed * dt;
+
+    // Once enemies are off the canvas, they reappear randomly with different speeds
+    if (this.x > 510) {
+        this.x = -50;
```
AMashoshyna: Numbers without clear meaning put directly into code are called "magic numbers" and are considered bad practice, because they make code hard to read. Please put the numbers into appropriately named constants.

submissions/denislukianenko/js-oop-frogger/js/app-ref.js
```diff
@@ -0,0 +1,125 @@
...
+        this.speed = 100 + Math.floor(Math.random() * 222);
+    };
+
+    // Checks for collisions between the player and the enemies
+    if (player.x < this.x + 102 &&
+        player.x + 102 > this.x &&
```
AMashoshyna: This condition is too long to keep it inside an `if` statement. Please put it into a constant or a named function. `isCollision` maybe?

submissions/denislukianenko/js-oop-frogger/js/app-ref.js
```diff
@@ -0,0 +1,125 @@
+// Enemies our player must avoid
+var Enemy = function (x, y, speed) {
```
AMashoshyna: `var` keyword is technically not deprecated, but please try to use `let`/`const` instead. `let`s and `const`s behave more predictably.

AMashoshyna:
Great job! Please check my comments.

AMashoshyna:
Oh, right, my mistake actually.
This version looks much better! The only thing I would like to fix is function `checkCollision`. Take a closer look at the condition. This condition evaluates to a boolean value. This means, you can return it directly:
```
function checkCollision(enemy) {
  return player.x < enemy.x + enemyBoundsX &&
    player.x + enemyBoundsX > enemy.x &&
    player.y < enemy.y + enemyBoundsY &&
    enemyBoundsY + player.y > enemy.y;
}
```
But the PR is good, I will merge it.

PR #89, opened 2018-12-11
Add js-post-oop (Edited Tiny JS World)

denislukianenko:
### A Tiny JS World (Post-oop)

[Demo](https://denislukianenko.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/denislukianenko/a-tiny-JS-world)

Please, review.


OleksiyRudenko:
@denislukianenko Thank you for the submission.

Some feedback from demo only review:
1. Animals do not report any limbs, which is required.
image

2. There is no man reported.

3. Alya reports friends. Yet there are no inhabitants named George, Denis, Vlad.
Please, note that if friend lists are supported those shouldn't be just arrays of strings, but rather lists of existent inhabitants.
image

4. Please, use single `print` per inhabitant. Hint: subclass method can call `super.methodName()` to call the method from superclass that it overrides.

Please, also check yourself against the following points:
 * [ ] Implement a base class to inherit from
 * [ ] Employ default parameters
 * [ ] Each species is represented with its own class
 * [ ] No need to specify species at instantiation
 * [ ] Classes for species that do not have hands by natural design
       do not consequently have `hands` or any equivalent property
 * [ ] All inhabitants are stored in a container (array or object)
 * [ ] JS native features are intensively employed
 * [ ] [OOP, SOLID and DRY](https://github.com/OleksiyRudenko/a-tiny-JS-world/blob/master/README.md#learn-on-your-own)
       principles are intensively employed
 
Optional level up (not required to implement):
 * [ ] Friends list is a list of objects refs rather than names (strings)
 * [ ] Cat-woman class is built employing composition rather than inheritance only
 
[ES6 classes cheat-sheet](https://gist.github.com/OleksiyRudenko/672d39b08d9d0da4e179aca49876c58b)

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,93 @@
+// ======== CLASSES ==========
+function Creature(name, gender, saying, legs) {
+  this.name = name ? name : "Creature";
+  this.gender = gender ? gender : "No gender";
+  this.saying = saying ? saying : "Que-qa";
+  this.legs = legs ? legs : 4;
+}
```
OleksiyRudenko: Species is a property that is common for all creatures.
Please, make it a property of this class.
Sub-classes may override any property.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,93 @@
+// ======== CLASSES ==========
+function Creature(name, gender, saying, legs) {
+  this.name = name ? name : "Creature";
+  this.gender = gender ? gender : "No gender";
+  this.saying = saying ? saying : "Que-qa";
+  this.legs = legs ? legs : 4;
```
OleksiyRudenko: Cross-checks are always useful. Please, also use default parameter values.
You may employ constants to be referred to in both default param values and cross-checks in constructor body.
Here and in other classes below as well.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,93 @@
...
+    "  Hands: " +
+    this.hands +
+    "<br>" +
+    "  Friends: " +
+    this.friends
+      .map(function(element) {
+        return element.name;
+      })
+      .join(", ");
+  print(this.printStr);
```
OleksiyRudenko: Will you find a method to benefit from existing `Creature.prototype.printInfo` i.e. re-using it and supplementing with info specific to a Human?
You are not required to fix it though.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,93 @@
...
+};
+//-------------------------------------------------------------------------------------
+function CatWoman(name, gender, saying, legs, hands, friends) {
+  Human.apply(this, arguments);
+  this.species = "cat-woman";
+  this.saying = cat.saying;
```
OleksiyRudenko: Classes are agnostic of (1) instances, especially of instances of other classes, (2) global variables (only exceptions are global constants and library functions and containers like `print`).
Please, fix.
Default sayings for cats may help.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,93 @@
...
+let catWoman = new CatWoman("Julia", "female", "", 2, 2, [woman, man, dog]);
+
+let allInhabitants = [];
+allInhabitants.push(dog, cat, woman, man, catWoman);
+// ======= OUTPUT =========
+dog.printInfo();
+cat.printInfo();
+woman.printInfo();
+man.printInfo();
+catWoman.printInfo();
```
OleksiyRudenko: How about `allInhabitants.forEach()`? You may have 1000's of.

OleksiyRudenko:
@denislukianenko Please, resolve comments you believe you've fixed.
That's how reviewer understands what is presumably done, not looking into the code itself. It also gives some ideas on whether latest commit contains all fixes or is a partial fix.
Thanks.

denislukianenko:
@OleksiyRudenko Thanks for a tip. Did it.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,99 @@
...
+};
+Creature.prototype.printInfo = function() {
+  this.formPrintString();
+  print(this.printStr);
+};
+
+//-----------------------------------------------------------------------------------
+function Cat(name, gender, saying, legs) {
+  Creature.apply(this, arguments);
+  this.species = "cat";
```
OleksiyRudenko: Since species is a property of the base class, please, pass it to `Creature.apply`.
Here and in other classes below.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,99 @@
...
+    "</span><br>" +
+    "  Gender: " +
+    this.gender +
+    "<br>" +
+    "  Says: " +
+    this.saying +
+    "<br>" +
+    "  Legs: " +
+    this.legs;
+};
```
OleksiyRudenko: The above creates a new property, freezing it. Not a really good approach as at some point (e.g. after giving a different name) other properties values may differ from what is "frozen" in `.printStr`.
Please, make this method returning a freshly formed string.
Another optimization: use [template strings](https://developers.google.com/web/updates/2015/01/ES6-Template-Strings).

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,99 @@
...
+Creature.prototype.printInfo = function() {
+  this.formPrintString();
+  print(this.printStr);
+};
```
OleksiyRudenko: If `.formPrintString()` returns a string then it can be used as an argument for `print` and also makes things more obvious than these currently are. It is not obvious that `print(...` instruction **requires** `this.formPrintString()` to be called before.

Another point - class must be agnostic of presentation media. Please, use `print` outside class definition, at reporting time.

So, basically you just need a single method that returns printable content.

Please, fix.

BTW, `formPrintString` is also very specific as to the media - it contains HTML markup. However, you are not required to fix this.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,99 @@
...
+  Creature.apply(this, arguments);
+  this.species = "human";
+  this.hands = hands ? hands : 2;
+  this.friends = friends ? friends : [];
+}
+Human.prototype = Object.create(Creature.prototype);
+Human.prototype.constructor = Human;
+Human.prototype.printInfo = function() {
+  Creature.prototype.formPrintString.apply(this);
+  this.printStr +=
```
OleksiyRudenko: Here you would call `formPrintString` from Creature and supplement it with some extra info, returning enriched string. Try finding this method via prototype chain.
Would be great if you've fixed this.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,99 @@
...
+    this.friends
+      .map(function(element) {
+        return element.name;
+      })
+      .join(", ");
+  print(this.printStr);
+};
+//-------------------------------------------------------------------------------------
+function CatWoman(name, gender, saying, legs, hands, friends) {
+  Human.apply(this, arguments);
```
OleksiyRudenko: This case would be better implemented using composition.

OleksiyRudenko:
@denislukianenko Why did you merge `master` into your feature branch? Please, explain.

Please, resolve conversations that you have already processed and/or give to know it is done in any other way. This gives a reviewer an idea of your current progress.

submissions/denislukianenko/js-post-oop/index.js
denislukianenko: Method is deleted. Now there is only formPrintString, which only returns string.

submissions/denislukianenko/js-post-oop/index.js
denislukianenko: Now there's only `formPrintString` method. In Human it calls Creature's version of itself, saves string into temp variable (no property creation here), concatenates saved string with additional info, available only in Human objects, and returns this enriched string. In print section we just pass method into print function, like `print(woman.formPrintString())`

submissions/denislukianenko/js-post-oop/index.js
denislukianenko: I would prefer not to change this

denislukianenko:
@OleksiyRudenko I've been updating this branch in such way:
(on js-post-oop branch)
– `git pull upstream master`
– paste index.js from my working folder in another directory into my homework folder
– commit and `push origin js-post-oop`

Yeah, doing so on js-post-oop creates merge but looks like shortest way, considering I have to Ctrl+V my file to homework folder anyway

OleksiyRudenko:
@denislukianenko Thanks for explanation.
You really didn't need any other code from within this repo for this task, and you might copy-paste it from your other repo.
Merging serves different purposes and it is often entangles repo history that other contributors may find unnecessary and affecting their work. So must be used with care. Keep this in mind.

submissions/denislukianenko/js-post-oop/index.js
```diff
@@ -0,0 +1,94 @@
...
+Creature.prototype.formPrintString = function() {
+  return (
+    '<span style="font-weight: bold; font-size: 23px">' +
+    this.name +
+    "</span>" +
+    '<span style="color: gray"> ' +
+    this.species +
+    "</span><br>" +
+    "  Gender: " +
+    this.gender +
```
OleksiyRudenko: JS template strings make code even more readable and lightweight.
You are not required to fix this. Keep in mind and use.
You may be required to fix such things in other tasks.

PR #91, opened 2018-12-12
JS DOM API task

rrrds:
### JS DOM API

[Demo](https://rrrds.github.io/kottans-hw-dom/) |
[Code base](https://github.com/rrrds/kottans-hw-dom)

Please, review.


submissions/rrrds/js-dom/js/main.js
```diff
@@ -0,0 +1,94 @@
...
+  content.dataset.indexNumber = idx;
+
+  main.appendChild(content);
+});
+
+navigation.appendChild(ul);
+
+document
+  .querySelector('.nav__item')
+  .dispatchEvent(new MouseEvent('click', { view: window, bubbles: true }));
```
zonzujiro: What for?

rrrds: To trigger first menus' item and set initial content view

zonzujiro: Why not make this with an initial state at HTML file?

rrrds: Set initial state while adding new content to page

PR #94, opened 2018-12-13
memoryPairGame_linkqwd

linkqwd:
### memoryPairGame

[Demo](https://linkqwd.github.io/piarGame/) |
[Code base](https://github.com/linkqwd/piarGame/blob/master/js/main.js)

### DIFF
--- a/submissions/linkqwd/task-10_memoryPairGame/main.js
+++ b/submissions/linkqwd/task-10_memoryPairGame/main.js
@@ -1,0 +1,117 @@
+const pairGame = function () {
+    const settings = {
+        cardContent: 'images', // colors or images
+        numberOfCardsToGuess: 2,
+        numberOfCardPairs: 10,
+        prewievDelay: 3000
+    }
+
+    let openedCardsCounter = settings.numberOfCardsToGuess;
+    let pairCounter = 0;
+    let scores = 0;
+
+    let DOM = {
+        card: 'js-card',
+        cardsHolder: 'js-cards-holder',
+        openedCard: 'js-card-opened',
+        noClickCard: 'js-card-no-click',
+        guessedCards: 'js-card-guessed',
+        scores: 'js-scores',
+        scoresSelector: document.querySelector('.js-scores')
+    }
+
+    let cardsResources = {
+        cardsColors: ['#e980a0', '#5880b5', '#a8804b', '#0080ff', '#dc8057', '#008000', '#ff7019', '#ca2b2c', '#edf906', '#26f906', '#f906b5', '#ffffff', '#2f1e3b', '#c700ff', '#0bd5cb', '#ff0700', '#2800ff', '#88b5a5'],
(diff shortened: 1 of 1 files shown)


submissions/linkqwd/task-10_memoryPairGame/main.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+        cards.sort(() => 0.5 - Math.random());
+        document.querySelector(`.${DOM.cardsHolder}`).insertAdjacentHTML('beforeend', cards.join(''));
+    }
+
+    const makePreview = () => {
+        toggleClass(DOM.noClickCard);
+        setTimeout(() => toggleClass(DOM.openedCard), 500);
+        setTimeout(() => toggleClass(DOM.openedCard), settings.prewievDelay + 500);
+        setTimeout(() => toggleClass(DOM.noClickCard), settings.prewievDelay + 1000);
```
zonzujiro: Magic numbers are bad - don't use it.
Also, to chain these events, you can use promises.

submissions/linkqwd/task-10_memoryPairGame/main.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    const updateScores = () => {
+        DOM.scoresSelector.textContent = scores;
+    }
+
+    const checkPair = () => {
+        const openedCards = document.querySelectorAll(`.${DOM.openedCard}`);
+        let IDsOfOpenedCards = [];
```
zonzujiro: Only constructors can be called from uppercase

submissions/linkqwd/task-10_memoryPairGame/main.js
```diff
@@ -0,0 +1,117 @@
...
+
+        openedCards.forEach(el => IDsOfOpenedCards.push(+el.dataset.id));
```
zonzujiro: You can do this with `Array#map`

linkqwd:
Thanks for review.
All done

zonzujiro:
> Thanks for review.
> All done

@linkqwd pls, resolve outdated discussions

submissions/linkqwd/task-10_memoryPairGame/main.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    makePreview(settings.animationDuration, () => toggleClass(DOM.openedCard))
+    .then(() => makePreview(settings.prewievDelay, () => toggleClass(DOM.openedCard)))
+    .then(() => makePreview(settings.animationDuration, () => toggleClass(DOM.noClickCard)));
+    
+    function checkPair () {
+        const openedCards = document.querySelectorAll(`.${DOM.openedCard}`);
+        let iDsOfOpenedCards = [...openedCards].map((el) => +el.dataset.id);
+        let isIDsEqual = iDsOfOpenedCards.reduce((stack, value) => (stack === value) ? stack = value : stack = 0);
```
zonzujiro: If you need to compare two id's - just compare two id's. No need in `reduce`, `map`, etc. But this is on future.

linkqwd: The point of this is that there can be more than two id's.
In game options it is possible to specify how many cards would be in one pair `numberOfCardsToGuess: 4`
So I had to use some solution like this.

PR #98, opened 2018-12-15
DOM API task

leonovoleksii:
### Document Object Model

[Demo](https://leonovoleksii.github.io/dom-js/) |
[Code base](https://github.com/leonovoleksii/dom-js)

Please, review.


AMashoshyna:
Very nice!
Let's see, what can be improved.
1. Generally it is better to keep your data and representation separate. You have put element names in `content` variable. Normally, this data would come from server and server does not need to know about your html structure. Please give the content items some generic names (title, image, description etc.). 
Bonus points: This is not required in this task, but it is better if you don't put anything in your html, but create all elements with JavaScript instead (imagine that one day you will have 5 menu items instead of 4; Hard-coding data in html is very inflexible). 
2. Use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Multi-line_strings) for multi-line strings (no `+` needed)
3. In line 55 you iterate the properties of an object and rely on their order. This is dengerous, because the order of properties in  JavaScript objects is not guaranteed. You can get your data as `{ p: 'Lorem ipsum', 'img': 'img/html.png', 'h1': 'HTML'}` and your layout would break. I suggest that you make one function to create chosen element (instead of `createTextElem` and `createImgElem`), then you can control the order in which the html elements are added. Try to keep this function outside of the loop (pass everything needed as arguments).

AMashoshyna:
Well done!

PR #100, opened 2018-12-17
OOP exercise

beta-version-profile:
### OOP exercise

[Demo](http://reminiscent-sail.surge.sh) |
[Code base](https://github.com/beta-version-profile/frontend-2019-homeworks/tree/oop_exercise/submissions/beta-version-profile/oop_exercies)

Please, review.

### DIFF
--- a/submissions/beta-version-profile/oop_exercies/index.js
+++ b/submissions/beta-version-profile/oop_exercies/index.js
@@ -1,0 +1,66 @@
+class Parent {
+  constructor(name, gender) {
+    this.name = name
+    this.gender = gender
+  }
+}
+
+class Animal extends Parent {
+  constructor(name, gender) {
+    super(name, gender)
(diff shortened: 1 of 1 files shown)


AMashoshyna:
Hi @beta-version-profile! Great job!
I will accept the PR, but please note, that taking methods from class instance (line 44) is not ok. I see why this was done, but this is not canonical OOP approach.
What can be done:
- best solution - use object composition. [This video](https://youtu.be/wfMtDGfHWpA) will give you the idea
- possible solution: create a Cat instance, pass it to CatWoman constructor, save it as CatWoman instance property and forward calls to `say` to this instance.

You are not required to fix it, but feel free to do it if you are interested. This is a rather advanced task.

beta-version-profile:
Thank you. I'll think about it. 
Inheritance is not the best thing in js)

PR #106, opened 2018-12-18
Add js-dom homework

webdevagent:

### dom-api-task

[Demo](https://webdevagent.github.io/) |
[Code base](https://github.com/webdevagent/frontend-2019-homeworks/tree/dom-api-task/submissions/webdevagent/js-dom)

Please, review.


submissions/webdevagent/js-dom/script.js
```diff
@@ -0,0 +1,23 @@
...
+      tab.descriptionItem.style.display = 'inline-block';
+    } else {
+      descriptions[i].style.display = 'none';
+    }
+  }
+};
+for (let i = 0; i < listItems.length; i++) {
+  listItems[i].descriptionItem = descriptions[i];
+  listItems[i].addEventListener('click', openTab);
+};
```
zonzujiro: Couple minor comments:

1. If you will not reassign a variable, use `const` instead of `let`. As an example, lambda functions are always declared with `const`
2. You should use array methods instead of loops because methods are declarative.

submissions/webdevagent/js-dom/script.js
```diff
@@ -0,0 +1,23 @@
...
+  for (let i = 0; i < listItems.length; i++) {
+    if (listItems[i] != tab) {
+      listItems[i].style.backgroundColor = '#80dfff';
+    } else {
+      tab.style.backgroundColor = 'white';
+    }
+  }
+  for (let i = 0; i < descriptions.length; i++) {
+    if (descriptions[i] == tab.descriptionItem) {
+      tab.descriptionItem.style.display = 'inline-block';
```
zonzujiro: Don't use inline style changing. It is a bad approach for several reasons: 

1. First of all, a browser will apply such styling for each line separately/ Which means, that every such line of code will be a reason for running of calculations of a page and for drawing it,  so you can receive a performance problem. Read about reflow and repaint.  
2. This is an imperative way, you need to write declarative and describe what your code does, not how. This will make your code shorter and easier to maintain. 
3. Reuse of code. Saying, you will need to rotate some other stuff - you will add a similar line to another part of an application. Which is not right because of DRY. 
4. JS for logic, CSS - for styling. 

So, replace this line with classes. You can use `classList` to manipulate them.

submissions/webdevagent/js-dom/script.js
```diff
@@ -0,0 +1,23 @@
+let listItems = document.querySelectorAll('.listItem');
+let descriptions = document.querySelectorAll('.description');
+let openTab = (evt) => {
+  let tab = evt.target;
```
zonzujiro: You can write like this:

```javascript
const openTab = ({ target }) => {...
```

And you don't need to create an additional variable just for getting a nested property.

webdevagent:
Thank's zonzujiro!
I  have been fixed the common mistakes, replaced let to const, loops to array methods.
Also, I removed inline style changing, and add instead classList.add() and classList.remove() methods.
And I simplified code with DRY method, as you mentioned before.

zonzujiro:
@webdevagent pls, resolve outdated discussions

PR #112, opened 2018-12-20
js-dom

RomanovAleksander:
##JS DOM

[Demo](https://romanovaleksander.github.io/js-dom/) |
[Code base](https://github.com/RomanovAleksander/kottans-frontend/tree/master/task_js_dom)

Please, review.


RomanovAleksander:
Do I need to do the task as in the example(photo) or is this also correct?

AMashoshyna:
@RomanovAleksander no, the design is completely up to you! No need to copy the example.

submissions/RomanovAleksander/js-dom/index.js
```diff
@@ -0,0 +1,87 @@
...
+        text: 'AngularJS (also written as Angular.js) is a JavaScript-based open-source front-end web application framework mainly maintained by Google and by a community of individuals and corporations to address many of the challenges encountered in developing single-page applications. The JavaScript components complement Apache Cordova, a framework used for developing cross-platform mobile apps.',
+        img: './assets/angular.svg'
+    },
+    {
+        title: 'Vue',
+        text: 'Vue.js is an open-source JavaScript framework for building user interfaces and single-page applications.' +
+        'Vue.js features an incrementally adoptable architecture that focuses on declarative rendering and component composition. Advanced features required for complex applications such as routing, state management and build tooling are offered via officially maintained supporting libraries and packages.',
+        img: 'assets/vue.svg'
+    }
+]);
```
AMashoshyna: Please add empty lines at the end of files.

submissions/RomanovAleksander/js-dom/index.js
```diff
@@ -0,0 +1,87 @@
...
+                item.classList.remove('item_active');
+            }
+        )
+
+    }
+
+    toggle() {
+        this.itemArray.forEach(item => {
+            item.addEventListener('click', () => {
+                    this.off();
```
AMashoshyna: Please use event delegation: adding event listeners in a loop is very ineffective. Add one event listener to containing element.

submissions/RomanovAleksander/js-dom/index.js
```diff
@@ -0,0 +1,87 @@
...
+const MENU_ACTIVE_CLASS = `${MENU_GENERAL_CLASS}_active`;
+const MENU_TEXT = `text`;
+const MENU_IMG_CLASS = `${MENU_GENERAL_CLASS}__img`;
+
+class Menu {
+    constructor(targetElement, array) {
+        this.targetEl = targetElement;
+        this.array = array;
+        this.itemArray = [];
+        this.render();
```
AMashoshyna: I call `render` here and then once for every element in the array, so you end up with 4 items in `this.itemArray` instead of 3.

submissions/RomanovAleksander/js-dom/index.js
```diff
@@ -0,0 +1,87 @@
...
+        this.listElement.appendChild(this.img);
+        this.listElement.appendChild(this.text);
+
+        this.itemArray.push(this.item);
+    }
+
+    renderList() {
+        for (let i = 0; i < this.array.length; i++) {
+            this.render();
+            this.targetEl.appendChild(this.item);
```
AMashoshyna: Changing DOM in a loop is very expensive in terms of performance. It is better to create a wrapper element, add all items and then append this element into the DOM. For example, you can use [document.createDocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment) to create a wrapper.

submissions/RomanovAleksander/js-dom/index.js
RomanovAleksander: img: 'assets/vue.svg ' should i do that?

AMashoshyna: image

No, after line 87. There is a red arrow to indicate that the newline is missing

RomanovAleksander: Ok, thanks

submissions/RomanovAleksander/js-dom/index.js
RomanovAleksander: I corrected all errors except this one. I did as in the example, but I can't display all elements. Сan you look at my code?

RomanovAleksander: Done

AMashoshyna: Your solution is fine, except of `renderList` method. In line 42 you attach the element to the DOM, and after you change attributes of element's children. It is better to make all adjustments before you attach the element.

Another thing that I have missed: in `toggle` method you provide a callback for `click` event listener and refer to `event` variable, but you don't pass it to the callback. It works in Chrome, because Chrome has global `event` variable, but in Firefox this code causes an error.

RomanovAleksander: Now it works everywhere)

AMashoshyna:
Well done!

PR #119, opened 2018-12-21
add a-tiny-JS-world task

Aldegid:
### A Tiny JS World

[Demo](https://aldegid.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Aldegid/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/Aldegid/a-tiny-js-world/index.js
+++ b/submissions/Aldegid/a-tiny-js-world/index.js
@@ -7,7 +7,7 @@
 
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
-function outCreature(obj) {
+function creature(obj) {
   var msg = [];
   for (let key of Object.keys(obj)) {
     msg.push(obj[key]);
@@ -59,11 +59,6 @@ const catWoman = {
   hands: 2,
   saying: woman.saying + " and sometimes " + cat.saying 
 };
-
-
-
-
-
 // ======== OUTPUT ========
 /* Use print(message) for output.
    Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
@@ -72,18 +67,8 @@ const catWoman = {
    However, please, REFRAIN from improving visuals at least until your code is reviewed
    so code reviewers might focus on a single file that is index.js.
    */
-  print(outCreature(dog));
-  print(outCreature(cat));
-  print(outCreature(man));
-  print(outCreature(woman));
-  print(outCreature(catWoman));
-
-/* Print examples:
-   print('ABC');
-   print('<strong>ABC</strong>');
-   print('<strong>ABC</strong>', 'div');
-
-   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
-   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
-   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
-   */
+  print(creature(dog));
+  print(creature(cat));
+  print(creature(man));
+  print(creature(woman));
+  print(creature(catWoman));
--- a/submissions/Aldegid/a-tiny-js-world/index.js
+++ b/submissions/Aldegid/a-tiny-js-world/index.js
@@ -59,6 +59,7 @@ const catWoman = {
   hands: 2,
   saying: woman.saying + " and sometimes " + cat.saying 
 };
+
 // ======== OUTPUT ========
 /* Use print(message) for output.
    Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
@@ -72,3 +73,4 @@ const catWoman = {
   print(creature(man));
   print(creature(woman));
   print(creature(catWoman));
+  
--- a/submissions/Aldegid/a-tiny-js-world/index.js
+++ b/submissions/Aldegid/a-tiny-js-world/index.js
@@ -1,0 +1,89 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+function outCreature(obj) {
+  var msg = [];
+  for (let key of Object.keys(obj)) {
+    msg.push(obj[key]);
+  }  
+  return  `Hello i am a <strong>${msg[0]}</strong>. My name is <strong>${msg[1]}</strong> and i am a ${msg[2]}. I have <strong>${msg[3]}</strong> legs and <strong>${msg[4]}</strong> hands and i saying: <em>${msg[5]}</em><hr>`;  
+}
+
+const dog = {
+  species: 'dog',
+  name: 'Toby',
+  gender: 'male',
+  legs: 4,
+  hands: 0,
+  saying: 'woof-woof!'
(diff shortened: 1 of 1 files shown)


submissions/Aldegid/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,89 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+function outCreature(obj) {
```
OleksiyRudenko: By convention function names denote actions so they should start with verbs.

submissions/Aldegid/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+/* Print examples:
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
```
OleksiyRudenko: Tweak your code editor so it ensures an empty line at the file end.

OleksiyRudenko:
@Aldegid Please, resolve conversations that you have processed.

submissions/Aldegid/a-tiny-js-world/index.js
Aldegid: i tied to set  "files.insertFinalNewline": true in my VS Code editor, but no result.

submissions/Aldegid/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,76 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+function creature(obj) {
```
OleksiyRudenko: "Creature" is a noun. We need a verb here. Just describe what function does.
E.g. `buildCreatureSelfPresentationString`. Not really a good example, too verbose. Just to give you an idea.

submissions/Aldegid/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,76 @@
...
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+  print(creature(dog));
+  print(creature(cat));
+  print(creature(man));
+  print(creature(woman));
+  print(creature(catWoman));
+
```
OleksiyRudenko: Looks like you preserve Windows specific end of line. You may need to change git settings.
There is an option to convert newlines to Linux style at commit and back to OS specific endings at checkout.

submissions/Aldegid/a-tiny-js-world/index.js
OleksiyRudenko: It doesn't create any inhabitants.
Can you please describe what it does? Plain human language, please.

Aldegid: I understood my mistake. When it came to renaming, I forgot what the function itself does. In this case, the function lists the properties (name, gender, etc.) of the inhabitant (represented in the code as object(s)) and outputs them to a string. So the function should be called like outputCreatureMessage

Aldegid: outCreatureMsg is corrent abbreviated?

OleksiyRudenko: Much better now )
You don't need abbreviation. And I would use `make` or `create` instead of `output` as `output` used as noun to often. Besides it doesn't do any output [to user interface].

submissions/Aldegid/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,67 @@
...
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+function makeCreatureMessage(obj) {
+  var msg = [];
+  for (let key of Object.keys(obj)) {
+    msg.push(obj[key]);
+  }
```
OleksiyRudenko: `Object.values()` does exactly what 4 lines of code above do.
You might use it. The only problem is that order of property keys and therefore of values as well is not guaranteed. If another contributor creates an object with properties in different order current approach will fail.
Please, use explicitly listed key names and map those to values.
You may skip this at this point but directly refer `obj` properties in code fragment below.

Aldegid: I rewrote the function using direct object key access. With this approach, problems with properties order really should not be. Thanks for the recommendation

OleksiyRudenko:
@Aldegid Did you finish?
There are several methods to let a code reviewer know you have finished an iteration:
- resolve conversations you believe you have properly processed
- let know explicitly
- respond to questions and inquiries (like "Please, confirm all items above are positive" above)

Communication is a key.

submissions/Aldegid/a-tiny-js-world/index.js
Aldegid: I implemented all the recommendations. Please review my final version of the code. And thank you for the recommendation and valuable experience of dialing via github. T

Aldegid:
I implemented all the recommendations. Please review my final version of the code. And thank you for the recommendation and valuable experience of dialing via github.

submissions/Aldegid/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,63 @@
...
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+function makeCreatureMessage(obj) {
+  return  `Hello i am a <strong>${obj.species}</strong>. My name is <strong>${obj.name}</strong> and i am a ${obj.gender}. I have <strong>${obj.legs}</strong> legs and <strong>${obj.hands}</strong> hands and i saying: <em>${obj.saying}</em><hr>`;
```
OleksiyRudenko: JS template strings can be multiline effectively keeping code readable.

submissions/Aldegid/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,63 @@
...
+const catWoman = {
+  species: cat.species + "-" + woman.species,
+  name: "Selina Kyle",
+  gender: "female",
+  legs: 2,
+  hands: 2,
+  saying: woman.saying + " and sometimes " + cat.saying
+};
+
+  print(makeCreatureMessage(dog));
```
OleksiyRudenko: No need for leading spaces.
Please, read about JS code formatting conventions. It is important as we write code for other people firsthand. You will be required to follow conventions.

submissions/Aldegid/a-tiny-js-world/index.js
Aldegid: thx accapted

PR #128, opened 2018-12-22
side-menu is done

AlinaLadybug:

### Document Object Model

[Demo](https://alinaladybug.github.io/DOM-Task/) |
[Code base](https://github.com/AlinaLadybug/DOM-Task)

Please, review.


submissions/AlinaLadybug/DOM/side-menu/js/script.js
```diff
@@ -0,0 +1,52 @@
+document.addEventListener("DOMContentLoaded", ready);
+
+//start function
+function ready() {
+    var contents = document.getElementById('content-wrapper').children;
+    
+    var li_elements = Array.from(document.getElementsByClassName('menu')[0].children);
```
zonzujiro: Ehm? https://github.com/kottans/frontend-2019-homeworks/pull/128#discussion_r243746284 🤔 

Can I help you with replacing this part of the code?

submissions/AlinaLadybug/DOM/side-menu/js/script.js
```diff
@@ -0,0 +1,52 @@
...
+    li_elements.forEach(element => {
+        element.addEventListener("mouseover", function () { hover(element) });
+        element.addEventListener("mouseout", function () { unhover(element) });
+        element.addEventListener("click", function () { changeContent(element) });
```
zonzujiro: You should use event delegation where it is possible. Adding of listener to each element – is not very effective. 

submissions/AlinaLadybug/DOM/side-menu/js/script.js
```diff
@@ -0,0 +1,52 @@
...
+}
+function unhover(element) {
+    if (element.classList.contains('active')) {
+        return;
+    }
+    var image = document.querySelector("."+element.classList[0]+" img");
+    var oldSrc = image.getAttribute('src');
+    image.setAttribute('src', oldSrc.replace('after', 'before'));
+}
+////////////////////////////
```
zonzujiro: Same with comments :) Can you remove it?

submissions/AlinaLadybug/DOM/side-menu/js/script.js
```diff
@@ -0,0 +1,52 @@
...
+
+
+function changeContent(element) {
+    //disable all contents
+    var contents = Array.from(document.getElementById('content-wrapper').children);
+    contents.forEach(element => {
+        element.classList.remove('active-content');
+    });
```
zonzujiro: Maybe better will be like this:

```javascript
document.querySelectorAll('.active-content')
```

Also, converting to an array is not obligatory, because `NodeList` has its own `forEach` method.

submissions/AlinaLadybug/DOM/side-menu/js/script.js
```diff
@@ -0,0 +1,52 @@
...
+
+    //disable active classes
+    var active = Array.from(document.getElementsByClassName('active'));
+    active.forEach(el=> {
+        var image = document.querySelector("."+el.classList[0]+" img");
```
zonzujiro: IMO, `el.classList[0]` this part looks like very fragile. What if someday your class list will have another value in the first position?

Take a look on [data-*](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) attributes.

submissions/AlinaLadybug/DOM/side-menu/js/script.js
AlinaLadybug: I try to employ event delegation to unordered list, but when some event happens, it applies only to li element.So when I try to hover li element it works(image is changed to 'after') but when hover image at all, it's changed again to 'before'.
```html
<ul class="menu">
    <li class="skype active">
      <img src="img/after/skype.svg" alt="skype" />
      <span class="tooltiptext">skype</span>
    </li>
```
```javascript
 var ul = document.querySelector('.menu');
 ul.addEventListener('mouseover',hover);
function hover(event) {
    if (!event.target.matches('li')) return;
    var element = event.target;
    var image = document.getElementsByClassName(element.className)[0].children[0];
    var oldSrc = image.getAttribute('src');
    image.setAttribute('src', oldSrc.replace('before', 'after'));
}
```

zonzujiro: I do not quite understand what you mean 😄 

Can you push your changes with this bug and I will look at it?

PR #131, opened 2018-12-23
Add first homework js-dom

Roman-Halenko:
### js-dom

[Demo](https://roman-halenko.github.io/dom-practical-task/) |
[Code base](https://github.com/Roman-Halenko/frontend-2019-homeworks/tree/js-dom-task/submissions/Roman-Halenko/js-dom)

Please, review.

### DIFF
--- a/submissions/Roman-Halenko/js-dom/css/style.css
+++ b/submissions/Roman-Halenko/js-dom/css/style.css
@@ -1,0 +1,170 @@
+html {
+  height: 100%;
+}
+
+body {
+  font-family: helvetica,sans-serif;
+  margin: 0;
+  height: 100%;
+}
+
+hr {
+  margin: 0px;
+}
+
+ul {
+  padding: 0px;
+}
+
+li {
+  list-style: none;
+  margin-top: 8px;
+  text-align: right;
+}
+
+a {
+  text-decoration: none;
+  color: #fffef6;
+  font-size: 24px;
+  transition: 0.2s ease;
+  font-weight: 200;
+}
+
+a:hover {
+  color: #1e3d46;
+}
+
+section {
+  display: flex;
+  height: 100%;
+}
+
+nav {
+  background: linear-gradient(to right, #336588, #4286a7f0);
+  flex-grow: 0;
+  flex-shrink: 0;
+  flex-basis: 20%;
+  padding: 40px 20px 20px 20px;
+  box-sizing: border-box;
+}
+
+nav.open {
+  -webkit-transform: translate(0, 0);
+  transform: translate(0, 0);
+}
+
+main {
+  padding: 40px 20px 20px 25px;
+  flex-basis: 80%;
+}
+
+span.btn-menu {
+  display: none;
+}
+
+@media screen  and (max-width: 750px) {
+  nav {
+    position: absolute;
+    width: 300px;
+    -webkit-transform: translate(-400px, 0);
+    transform: translate(-400px, 0);
+    transition: transform 0.5s ease;
+    height: 100%;
+  }
+
+  main {
+    flex-basis: 100%;
+  }
+
+  span.btn-menu {
+    box-sizing: border-box;
+    display: flex;
+    position: absolute;
+    height: 25px;
+    width: 25px;
+    border: 2px solid #b89a6e;
+    top: 25px;
+    left: 25px;
+    border-radius: 30%;
+    background-color: #b89a6e;
+    z-index: 1000;
+  }
+
+  div.content-box {
+    width: 100%;
+  }
+}
+
+.isActive {
+  color: #1e3d46;
+}
+
+.isActive::after {
+  content: '';
+  position: relative;
+  display: inline-block;
+  height: 0px;
+  width: 0px;
+  left: 20px;
+  bottom: 0px;
+  border-top: 10px solid transparent;
+  border-bottom: 10px solid transparent;
+  border-right:10px solid white;
+}
+
+.btn-menu::before, .btn-menu::after {
+  content: "";
+  border-top: 3px solid #ffffff;
+  width: 80%;
+  height: 0px;
+  position: absolute;
+  left: 9%;
+  border-radius: 1px;
+}
+
+.btn-menu::before {
+  top: 6px;
+}
+
+.btn-menu::after {
+  top: 13px;
+}
+
+@media screen  and (min-width: 751px) and (max-width: 1220px){
+  nav {
+    flex-basis: 30%;
+  }
+}
+
+.content {
+  flex-wrap: wrap;
+  display: flex;
+  align-items: start;
+}
+
+.title {
+  width: 100%;
+  font-size: 32px;
+}
+
+.description {
+  line-height: 1.5em;
+  width: 50%;
+}
+
+.circle {
+  width: 120px;
+  max-width: 100%;
+}
+
+.photo {
+  width: 450px;
+  max-width: 100%;
+}
+
+.content-box {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: start;
+  width: 50%;
+}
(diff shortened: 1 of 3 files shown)


submissions/Roman-Halenko/js-dom/main.js
```diff
@@ -0,0 +1,86 @@
...
+
+const menuItems = document.querySelectorAll('.menu-item');
+
+for (let i = 0; i < menuItems.length; i++) {
+  menuItems[i].addEventListener('click', function() {
+    ttl.innerText = titlesText[i];
+    dsc.innerText = descriptionsText[i];
+    sticker.setAttribute("src", stickersImgSrc[i]);
+    preview.setAttribute("src", prevImgSrc[i]);
+    menuItems.forEach(function(item) {item.classList.remove('isActive')});
```
IgorKurkov: you can get out this foreach to separate function because there you make more than one thing, but as we know - a function must do only one thing.

submissions/Roman-Halenko/js-dom/index.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en-Us" dir="ltr">
+  <head>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta charset="utf-8">
+    <title>Find your cap style</title>
+    <link rel="stylesheet" href="css/style.css">
+    <script type="text/javascript" async src="main.js"></script>
```
IgorKurkov: I think js script better be usable before `</body>` tag. CSS in head, js above </body>

zonzujiro:
@Roman-Halenko do you need some help with this task?

Roman-Halenko:
@zonzujiro Hi! I'm just starting to change code following the suggestions. Should I make a new one commit for each of them?

submissions/Roman-Halenko/js-dom/css/style.css
```diff
@@ -0,0 +1,170 @@
...
+  top: 6px;
+}
+
+.btn-menu::after {
+  top: 13px;
+}
+
+@media screen  and (min-width: 751px) and (max-width: 1220px){
+  nav {
+    flex-basis: 30%;
```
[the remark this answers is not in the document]
Roman-Halenko: Hi! Big thank for your attention to my task solution. I detect that second menu item (the longest one) wrapping in two lines between 751vw and 813vw. Is that what I should fix by changing a width of navigation bar/menu item?

IgorKurkov:
> @zonzujiro Hi! I'm just starting to change code following the suggestions. Should I make a new one commit for each of them?

no, you can do all the changes in any count of commits as you wish

Roman-Halenko:
@IgorKurkov, I've done some of the changes, but still not sure about function of event listener (separating)

submissions/Roman-Halenko/js-dom/main.js
```diff
@@ -0,0 +1,94 @@
...
+
+const dsc = document.createElement('p');
+dsc.className = 'description';
+dsc.innerText = data[0].description;
+cBox[0].appendChild(dsc);
+
+const sticker = document.createElement('img');
+sticker.setAttribute("src", data[0].stickerImg);
+sticker.className = 'circle';
+cBox[0].appendChild(sticker);
```
zonzujiro: Don't use constructions like `children[0]`, `parentNode`, `nextSibling`, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use `querySelector` or smth instead.

submissions/Roman-Halenko/js-dom/main.js
```diff
@@ -0,0 +1,94 @@
+const menuBtn = document.getElementById('btn-menu');
+const navBar = document.getElementById('navbar');
+const main = document.getElementsByTagName('main')[0];
```
zonzujiro: Better use `querySelector`, it will return one item.

submissions/Roman-Halenko/js-dom/main.js
```diff
@@ -0,0 +1,94 @@
...
+const menuList = document.querySelector('#navbar ul');
+
+// Mobile navigation
+menuBtn.addEventListener('click', e => {
+  navBar.classList.toggle('open');
+  e.stopPropagation();
+});
+
+main.addEventListener('click', () => {
+  if (navBar.className === "open") {
```
zonzujiro: Better use `classList.contains`, because one day you may change class names or add space, by mistake. Or smth else.

submissions/Roman-Halenko/js-dom/main.js
```diff
@@ -0,0 +1,94 @@
...
+preview.className = 'photo';
+cBox[1].appendChild(preview);
+
+function makeActive(trg, elements) {
+  elements.forEach(elem => {elem.classList.remove('isActive')});
+  trg.classList.add('isActive');
+}
+
+menuList.addEventListener('click', (ev) => {
+  for (let i = 0; i < menuItems.length; i++) {
```
zonzujiro: You should write in the declarative, not imperative way. 

Useful links: 

submissions/Roman-Halenko/js-dom/main.js
```diff
@@ -0,0 +1,94 @@
...
+
+const sticker = document.createElement('img');
+sticker.setAttribute("src", data[0].stickerImg);
+sticker.className = 'circle';
+cBox[0].appendChild(sticker);
+
+const preview = document.createElement('img');
+preview.setAttribute("src", data[0].prevImg);
+preview.className = 'photo';
+cBox[1].appendChild(preview);
```
zonzujiro: I am not sure about indexes here. How you think, can you replace it with something?

zonzujiro:
@Roman-Halenko pls, resolve outdated discussions

submissions/Roman-Halenko/js-dom/main.js
Roman-Halenko: I thought I could save memory using getElementsByTagName('main')[0] instead of querySelector.  Perhaps I get wrong with info from @IgorKurkov in second suggestion.

submissions/Roman-Halenko/js-dom/main.js
Roman-Halenko: Is that question only about cBox[1] or, also, getting data from the array data[0].prevImg?

Roman-Halenko:
@zonzujiro should I build more strict interdependence between menu items and appropriate content by providing ids to data objects or smth like that?

submissions/Roman-Halenko/js-dom/main.js
zonzujiro: Well... it is not about memory, but about performance and `querySelector` is slower, yes. But you don't have a lot of elements on the screen, therefore you will not see a difference between these two selectors. Except, that `querySelector` will look cleaner and will not create a `NodeList`.

Also, general rule - don't do premature optimization. Because it will get you into the land with code, which was overcomplicated without a reason. And this cause bugs, and with such code hard to work, etc.

submissions/Roman-Halenko/js-dom/main.js
zonzujiro: Yeah, you can extract elements from array with destructuring.

zonzujiro:
@Roman-Halenko it can be a good idea, yes.

submissions/Roman-Halenko/js-dom/main.js
Roman-Halenko: Accepted. Thank you :)

IgorKurkov:
> @IgorKurkov, I've done some of the changes, but still not sure about the function of event listener (separating)

I told you about single eventListener on the container element - you can read about event delegation and be more skilled in these cases. 
But I think you can improve this later, and focused now on other tasks so - I think LGTM now

IgorKurkov:
@zonzujiro ?

Roman-Halenko:
@zonzujiro, @IgorKurkov - thank a lot :)

PR #133, opened 2018-12-23
Complete dom-api-task

bugagashinka:
### dom-api-task

[Demo](https://bugagashinka.github.io/task_js_dom/) |
[Code base](https://github.com/bugagashinka/task_js_dom)

The code is submitted in a dedicated feature branch.
Please, review.

PR #136, opened 2018-12-23
Memory pair game

bugagashinka:
### memory-pair-game

[Demo](https://bugagashinka.github.io/memory-pair-game/) |
[Code base](https://github.com/bugagashinka/memory-pair-game)

Please, review.


OleksiyRudenko:
@bugagashinka Not a really good idea to remove another task files. This will remove those from the repo.
Please, fix issues with repo before proceeding with code review.
You may find [this](https://github.com/kottans/frontend-2019-homeworks/pull/url) helpful.

OleksiyRudenko:
@bugagashinka Good, thanks. Now we need PR having only relevant code base.
Did you have a chance to read the material following the link in my past post above?

bugagashinka:
Nope, link direct to page that doesn't exist. Can you clarify What you mean for 'relevant code base', without 'Frogger project' or ...

OleksiyRudenko:
@bugagashinka 
Here is the correct [link](https://gist.github.com/OleksiyRudenko/8b3ddb664308de0634b53c525e551d8b)

> Can you clarify What you mean for 'relevant code base', without 'Frogger project' or ...

This (and any either) PR should contain only relevant commits. In this particular case, only commits related to Memory Pair Game.

This target can achieved with proper branching.

bugagashinka:
> @bugagashinka
> Here is the correct [link](https://gist.github.com/OleksiyRudenko/8b3ddb664308de0634b53c525e551d8b)
> 
> > Can you clarify What you mean for 'relevant code base', without 'Frogger project' or ...
> 
> This (and any either) PR should contain only relevant commits. In this particular case, only commits related to Memory Pair Game.
> 
> This target can achieved with proper branching. Please, check the link above. Should you require any further clarifications, please, let me know.


Thanks, it really helped

OleksiyRudenko:
@bugagashinka Great, now your PR won't break any code out of the task scope.

OleksiyRudenko:
@bugagashinka Impression from demo: looks nice, the only confusing thing is that board gets reset pretty randomly, and also seems like upon board reset cards keep their places.
Is it some intended behaviour?

bugagashinka:
> @bugagashinka Impression from demo: looks nice, the only confusing thing is that board gets reset pretty randomly, and also seems like upon board reset cards keep their places.
> Is it some intended behaviour?

Thanks for positive feedback. 
Board always reset after 2 seconds on completion of the game, this intended behaviour.
What about card, may need to improve shuffle.

OleksiyRudenko:
@bugagashinka Note how the board resets while there are still 4 cards to play.

record_2018_12_25_21_40_37_463

bugagashinka:
@OleksiyRudenko Thanks, hope I fixed that. I think problem was that the card handle double click and program logic interpreted this as pair, also I fixed random order of cards.

submissions/bugagashinka/memory-pair-game/js/main.js
```diff
@@ -0,0 +1,106 @@
...
+      card.children[0].classList.add('front-disabled');
+    });
+    if (pairCount == cardImgArr.length) {
+      setTimeout(resetGame, WAIT_TIME_BEFORE_RESET);
+    }
+  }
+
+  function cardPairCheck() {
+    return (
+      flippedCardArr[0].children[1].children[0].src ==
```
zonzujiro: Don't use constructions like `children[0]`, `parentNode`, `nextSibling`, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use `querySelector` or smth instead.

submissions/bugagashinka/memory-pair-game/js/main.js
```diff
@@ -0,0 +1,106 @@
...
+  const gameBoard = doc.querySelector('.board');
+
+  gameBoard.addEventListener('click', event => {
+    let target = event.target.classList;
+    if (!target.contains('front-disabled') && target.contains('front')) {
+      flippCardBackEnd(event.target.parentElement);
+    }
+  });
+
+  gameBoard.addEventListener('transitionend', e => {
```
zonzujiro: 👍

submissions/bugagashinka/memory-pair-game/js/main.js
```diff
@@ -0,0 +1,106 @@
+(function(global) {
+  const MAX_FLIPP_CARDS = 2,
+    SHOW_CARDS_TIME = 1000,
+    WAIT_TIME_BEFORE_RESET = 2000;
+
+  const doc = global.document,
```
zonzujiro: You don't need this part, because your function is running inside the `window` object, so it will available to you.

submissions/bugagashinka/memory-pair-game/js/main.js
```diff
@@ -0,0 +1,106 @@
...
+      'images/3.jpg',
+      'images/4.jpg',
+      'images/5.jpg',
+      'images/6.jpg',
+    ];
+
+  const gameBoard = doc.querySelector('.board');
+
+  gameBoard.addEventListener('click', event => {
+    let target = event.target.classList;
```
zonzujiro: `classList` is an array with class names, and `target` is a DOM element. Pls, change naming.

Also, you can use destructuring and write like this `gameBoard.addEventListener('click', ({ target }) => `

submissions/bugagashinka/memory-pair-game/js/main.js
```diff
@@ -0,0 +1,106 @@
...
+    if (!target.contains('front-disabled') && target.contains('front')) {
+      flippCardBackEnd(event.target.parentElement);
```
zonzujiro: As was told, you rely on DOM structure. Don't.

submissions/bugagashinka/memory-pair-game/js/main.js
bugagashinka: Thanks for review and tips

submissions/bugagashinka/memory-pair-game/js/main.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  function cardPairCheck() {
+    return (
+      getElementBy('.back img', flippedCardArr[0]).src ==
+      getElementBy('.back img', flippedCardArr[1]).src
+    );
+  }
+
+  function getElementBy(selector, root) {
```
zonzujiro: IMO, redundant helper. Why you need it?

submissions/bugagashinka/memory-pair-game/js/main.js
```diff
@@ -0,0 +1,111 @@
...
+    return root.querySelector(selector);
+  }
+
+  const cardNodeArr = Array.prototype.slice.call(
```
zonzujiro: `Array.from`. But why you need a shallow copy of array?

submissions/bugagashinka/memory-pair-game/js/main.js
bugagashinka: I'm little bit confused, cuz in another my PR #133 I get another advice about DRY pattern, also I can make this helper function more univeral, for example:
```javascript
function getElementBy(selector, root=document) {
    return root.querySelector(selector);
}
```
Please give response before I start do any change. Thanks.

submissions/bugagashinka/memory-pair-game/js/main.js
bugagashinka: Thank, my bad

zonzujiro: > But why you need a shallow copy of array

@bugagashinka ^

submissions/bugagashinka/memory-pair-game/js/main.js
zonzujiro: `document.querySelector('.class')` is more dry then `getElementBy('.class', document)`? 🤔 How?

Maybe you can give a link to concrete comment? Because I see the only discussion regarding DRY and it is about the creation of elements.

zonzujiro: Creation of element - is several operations. Like create an element, add a class, event listeners, etc.

Selecting of element - is one operation. You just have replaced native API one-liner with your one-liner and ended with redundant function.

This is not DRY 😄 You understand why?

PR #138, opened 2018-12-23
Dom api task od

OlgaFrontend:
### [Demo](https://olgafrontend.github.io/kottans-frontend/task_js_dom/practice_js_dom/) |  [Code](https://github.com/OlgaFrontend/kottans-frontend/tree/master/task_js_dom/practice_js_dom)

Please, review.


OleksiyRudenko:
Please, edit the initial pr post that is created from template so its content reflects your actuals.

submissions/OlgaDubynka/js_dom/script.js
```diff
@@ -0,0 +1,46 @@
...
+
+logo.addEventListener('click', function() {
+  if(menuBtn.classList.contains('active')) {
+    menuBtn.classList.remove('active');
+  }
+  if(navList.classList.contains('active')) {
+    navList.classList.remove('active');
+  }
+  const elem = content.querySelector("img");
+  elem.remove();
```
AMashoshyna: `img` element is not always present in the DOM. If you click the logo before you open the first image or click it more than once, you will get an error in the console. 

image

It is better to check if the element in not null before calling any methods on it.

submissions/OlgaDubynka/js_dom/script.js
```diff
@@ -0,0 +1,46 @@
...
+  if (listItem.nodeName === 'LI') {
+    const indexOfActiveItem = getIndexOfActiveItem(listItem);
+    const img = filterData(indexOfActiveItem, data);
+    content.innerHTML = '<img src="'+ img + '" />'
+  }
+});
+
+function filterData(indexOfActiveItem, data) {
+  let img;
+  const arr = data.filter(item => {
```
AMashoshyna: It does not look like a good use case for `Array.prototype.filter`, which is meant to produce a new array of items matching the condition.
How about `Array.prototype.find`? Again, this method may return `undefined`. Check before accessing properties.

submissions/OlgaDubynka/js_dom/script.js
```diff
@@ -0,0 +1,46 @@
...
+    if (item.id === indexOfActiveItem) {
+      img = item.img;
+    };
+  });
+  return img;
+}
+
+function getIndexOfActiveItem(targetItem) {
+  const arr = [...menuItems];
+  const indexItem = arr.indexOf(targetItem);
```
AMashoshyna: No need to create a variable here, you can just return `arr.indexOf(targetItem)`

submissions/OlgaDubynka/js_dom/data.js
```diff
@@ -0,0 +1,22 @@
...
+    id: 2,
+    img: 'img/summer.jpg'
+  },
+
+  {
+    id: 3,
+    img: 'img/autumn.jpg'
+  },
+
+];
```
AMashoshyna: Please add newline at the end of every file and remove unnecessary empty lines.

zonzujiro:
@OlgaFrontend pls, resolve outdated discussions.

AMashoshyna:
Hi @OlgaFrontend! Great job, code looks much better!
The last thing to fix is `getImage` function. Here is your implementation:
```javascript
function getImage(indexOfActiveItem, data) {
    let img;
    const arr = data.find(item => {
        if (item.id === indexOfActiveItem) {
            img = item.img;
        }
    });
    return img;
}
```

Basically, you use Array.prototype.find (and previously Array.prototype.filter) to iterate the array, which is technically possible, but semantically incorrect and ineffective. Both Array.prototype.find and  Array.prototype.filter expect a function which returns a boolean value indicating whether current item matches the condition. So I would write it like this:
```javascript
function getImage(indexOfActiveItem, data) {
    const activeDataItem = data.find(item => item.id === indexOfActiveItem);
    if(activeDataItem) {
        return activeDataItem.img;
    }
    return '';
}
```
Let me know if you need more detailed explanation of the code above.

AMashoshyna:
Great job!

PR #139, opened 2018-12-24
add completed js-oop task

leonovoleksii:

### JS OOP Frogger Game

[Demo](https://leonovoleksii.github.io/frontend_nanodegree_arcade_game/) |
[Code base](https://github.com/leonovoleksii/frontend_nanodegree_arcade_game)

Please, review.


OleksiyRudenko:
@leonovoleksii Took a quick look. Looks consistent.
Please, check yourself against the following list. It might help you improving your code like defining constants etc.

 * Requirements re **OOP**:
 * Most common mistakes
         
OleksiyRudenko:
@leonovoleksii Do you need any support?

zonzujiro:
Looks like that task was abandoned by the author. I will close it. But we always can open it again.

PR #141, opened 2018-12-24
Task 7. A Tiny JS World

babayK0:
###Task 7. A Tiny JS World

[Demo](https://babayk0.github.io/a-tiny-JS-world/) 
 [Code base](https://github.com/babayK0/a-tiny-JS-world)

Please, review.


submissions/babayK0/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,91 @@
...
+dog.friends = `${cat.name} and ${man.name}`;
+cat.friends = `${dog.name} and ${woman.name}`;
+woman.friends = `${man.name} and ${cat.name}`;
+man.friends = `${woman.name} and ${dog.name}`;
+catwoman.friends = `${woman.name} and ${woman.friends}`
+
+function creatingOrganism(obj){
+   if(obj.friends!==''){
+      return `I am ${obj.species}-${obj.gender}. My name is ${obj.name}.I have ${obj.legs} legs and ${obj.hands} hands.My friends are ${obj.friends}.My favorite phrase is: <strong>${obj.saying}</strong>`;
+   }else{
```
OleksiyRudenko: If we add yet another property we'll be required to fix two very similar string compositions.

Please employ [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) to make code cleaner and [DRYer](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

submissions/babayK0/a-tiny-JS-world/index.js
```diff
@@ -59,14 +59,12 @@ man.friends = `${woman.name} and ${dog.name}`;
 catwoman.friends = `${woman.name} and ${woman.friends}`
 
 function creatingOrganism(obj){
-   if(obj.friends!==''){
-      return `I am ${obj.species}-${obj.gender}. My name is ${obj.name}.I have ${obj.legs} legs and ${obj.hands} hands.My friends are ${obj.friends}.My favorite phrase is: <strong>${obj.saying}</strong>`;
-   }else{
-      obj.friends = '...oh,i haven\'t friends =( ';
+   if(obj.friends ===''){
+   obj.friends = '...oh,i haven\'t friends =( ';
```
OleksiyRudenko: Not a really good solution.
Here is why:
When an object has friends `.friends` contains a string. When object doesn't have friends `.friends` contains a string. From app point of view there is no much difference in state between an object having friends and an object having no friends.

Try something like:
```
return `I am ${obj.species}....`
 + (obj.friends ? '...oh, i haven\'t friends' : `My friends are ${obj.friends}`)
 + `My favourite phrase is ...`;
```

By the way, when something is not defined use `undefined` instead of `''`, as the latter is a value just zero-long.

submissions/babayK0/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,86 @@
...
+   friends: undefined
+};
+dog.friends = `${cat.name} and ${man.name}`;
+cat.friends = `${dog.name} and ${woman.name}`;
+woman.friends = `${man.name} and ${cat.name}`;
+man.friends = `${woman.name} and ${dog.name}`;
+catwoman.friends = `${woman.name} and ${woman.friends}`
+
+function creatingOrganism(obj){
+      return `I am ${obj.species}-${obj.gender}. My name is ${obj.name}.I have ${obj.legs} legs and ${obj.hands} hands.`+(obj.friends==undefined ? '...oh, i haven\'t friends' : `My friends are ${obj.friends}`)+`.My favorite phrase is: <strong>${obj.saying}</strong>`;
```
OleksiyRudenko: It is recommended to split long strings. It is good doing so at operators. Also template strings can be multi-line.

submissions/babayK0/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,86 @@
...
+   saying: cat.saying,
+   friends: undefined
+};
+dog.friends = `${cat.name} and ${man.name}`;
+cat.friends = `${dog.name} and ${woman.name}`;
+woman.friends = `${man.name} and ${cat.name}`;
+man.friends = `${woman.name} and ${dog.name}`;
+catwoman.friends = `${woman.name} and ${woman.friends}`
+
+function creatingOrganism(obj){
```
OleksiyRudenko: Function name confuses as its body doesn't contain any instructions to create any organisms.
We write code for other people, firsthand. Please, keep this in mind.

babayK0:
@OleksiyRudenko  Ok,I understand you .I will try to fix it in next time.Thank you for review)

PR #144, opened 2018-12-25
Added js-dom task

prhokay:

### js-dom

[Demo](https://prhokay.github.io/) | 
[Code base](https://github.com/prhokay/prhokay.github.io)

Please, review.

### DIFF
--- a/submissions/prhokay/js-dom/index.html
+++ b/submissions/prhokay/js-dom/index.html
@@ -1,0 +1,20 @@
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Cities</title>
+    <link rel="stylesheet" href="css/style.css">
+</head>
+<body>
+    <div class="nav">
+      <button class="nav-link" data-city="London">London</button>
+      <button class="nav-link" data-city="Paris">Paris</button>
+      <button class="nav-link" data-city="Tokyo">Tokyo</button>
+    </div>
+    
+    <div id="main"></div>
+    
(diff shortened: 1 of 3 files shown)


submissions/prhokay/js-dom/main.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+  {
+    id: 'Tokyo',
+    text: 'Tokyo is the capital of Japan.',
+  },
+];
+
+const mainContent = document.getElementById('main');
+
+document.querySelectorAll('.nav-link').forEach(registerMenuLinkClickEventListener);
```
zonzujiro: You should use event delegation where it is possible. Adding of listener to each element – is not very effective. 

submissions/prhokay/js-dom/index.html
```diff
@@ -0,0 +1,20 @@
...
+      <button class="nav-link" data-city="London">London</button>
+      <button class="nav-link" data-city="Paris">Paris</button>
+      <button class="nav-link" data-city="Tokyo">Tokyo</button>
+    </div>
+    
+    <div id="main"></div>
+    
+    <script type="text/javascript" src="main.js"></script>
+</body>
+</html>
```
zonzujiro: This red sign is here because you need to add empty line in the end of the file ;)

submissions/prhokay/js-dom/main.js
```diff
@@ -0,0 +1,25 @@
+const data = [
```
zonzujiro: `data` is too general name. Can you maybe change it to something more concrete? `cities` maybe?

I understand, that for this amount of code `data` is the understandable name, but it still has a wide meaning.

zonzujiro:
Looks like that PR was abandoned by the author. Therefore I will close it but we always can open it again.

PR #148, opened 2018-12-25
friendsApp task ready for review

linkqwd:
### Friends application

[Demo](https://linkqwd.github.io/friendApp/) 
[Code base](https://github.com/linkqwd/friendApp/blob/master/js/main.js)

### DIFF
--- a/submissions/linkqwd/task_11_friendsApp/main.js
+++ b/submissions/linkqwd/task_11_friendsApp/main.js
@@ -1,7 +1,3 @@
-const settings = {
-    friendsCount: 50,
-}
-
 const DOM = {
     friendHolder: 'b-friends-list',
     friendCard: 'b-friends-list__friend-item',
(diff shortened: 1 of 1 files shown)


submissions/linkqwd/task_11_friendsApp/main.js
```diff
@@ -0,0 +1,100 @@
...
+
+let ucFirst = (str) => str.charAt(0).toUpperCase() + str.slice(1); // UpperCase For first letter
+
+function renderHTMLFriendsCards(friendsArray) {
+    let friendsHTMLCards = [];
+    for (let i = 0; i < friendsArray.length; i++) {
+        friendsHTMLCards.push(`<li class="${DOM.friendCard} ${!friendsArray[i].isPresentInList ? 'hidden' : ''}" id="${friendsArray[i].id}"><div class="b-friends-list__friend-item-wrapper"><div class="b-friends-list__img-wrapper"><img class="b-friends-list__image" src="${friendsArray[i].photo}" alt="test"></div><div class="b-friends-list__info-wrapper"><div class="b-friends-list__friend-name b-friends-list__info-row"><span>${ucFirst(friendsArray[i].name)} ${ucFirst(friendsArray[i].lastName)}</span></div><div class="b-friends-list__friend-age b-friends-list__info-row"><span>Age: ${friendsArray[i].age}</span></div><div class="b-friends-list__friend-age b-friends-list__info-row"><span>Gender: ${friendsArray[i].gender}</span></div><div class="b-friends-list__friend-phone b-friends-list__info-row"><span>Phone: ${friendsArray[i].cell}</span></div></div></div></li>`);
+    }
+
+    DOM.friendHolderSelector.innerHTML = friendsHTMLCards.join('');
```
IgorKurkov: its okay but keep in mind that innerHTML with strings is a more difficult operation for browser and it must first parse a string to DOM elements and after that insert them but you can prepare this yourself by `htmlFragments` and another DOM functions like `adjacent`'s and others `insert`s. Newer mind

submissions/linkqwd/task_11_friendsApp/main.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function filteringByName() {
+    friendsList.forEach((friend) => ~friend.fullName.indexOf(this.value.toLowerCase()) ? friend.showInList() : friend.removeFromList());
```
IgorKurkov: all great and indexOf very fast - you choose a good method, but indexOf and search have similar results but search more descriptive and receive regex which can be made on the fly like `friend.fullName.search(new RegExp(this.value, 'igm'))`, but its only a suggestion. See more in [regex docs](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/RegExp)

submissions/linkqwd/task_11_friendsApp/main.js
linkqwd: Thank you. I haven't thought that it might be harder for browser to parse string into DOM elements, then provide for browser already built Nodes. Good to keep in mind.
But for me this approach is so much more convenient, that I couldn't resist using it :)
In future I would probably change my habits, but if you may, not for now.

linkqwd:
Thank you for review. Good things which I missed.
I've tried to improve readability of code by your guidelines. Hope I got it correctly.
Placing <p> tag into <label> wasn't in my intentions, as well as keeping those cake images from first practical task. Just an accident :)

PR #150, opened 2018-12-25
Js dom

nazmariam:
### Document Object Model

[Demo](https://nazmariam.github.io/kottans/index.html) |
[Code base](https://github.com/nazmariam/nazmariam.github.io/tree/master/kottans)

First commit was mine, I forgot to config git at new repository.

Please, review.


AMashoshyna:
Great job! 
What can be improved:
- code formatting. Please check inconsistent indentations and empty lines
- manipulating DOM nodes is expensive in terms of performance. Try to reduce queries to DOM whenever possible. Example: you make query `document.querySelectorAll('li')` once, the iterate the result and repeat the query on every loop iteration. Remember, that the browser needs to traverse the whole DOM to find `li` elements. Modern browsers are optimized and fast, but still it is a relatively slow operation. Instead you can make the query, save result in a variable and work with it.
- adding event listeners in a loop is ineffective. Please use [event delegation](https://davidwalsh.name/event-delegate) instead: add one listener to container element.
- the layout seems to be incorrect (Google Chrome on Ubuntu):
image

nazmariam:
Hi! Thank you so much for review. Really there was a Udacity lesson about event delegation, I forgot about it. It shows the difference between theoretical knowledge and practice.

nazmariam:
> Great job!
> What can be improved:
> 
> * code formatting. Please check inconsistent indentations and empty lines
> * manipulating DOM nodes is expensive in terms of performance. Try to reduce queries to DOM whenever possible. Example: you make query `document.querySelectorAll('li')` once, the iterate the result and repeat the query on every loop iteration. Remember, that the browser needs to traverse the whole DOM to find `li` elements. Modern browsers are optimized and fast, but still it is a relatively slow operation. Instead you can make the query, save result in a variable and work with it.
> * adding event listeners in a loop is ineffective. Please use [event delegation](https://davidwalsh.name/event-delegate) instead: add one listener to container element.
> * the layout seems to be incorrect (Google Chrome on Ubuntu):
>   image

Done! Could you check, please?

AMashoshyna:
I see, that the demo looks great now (so the last issue is fixed), but I don't see new commit in the PR. Are you sure you have merged the right branch?

PR #151, opened 2018-12-25
add js-dom task

allykrupina:
### js-dom task

[Demo](https://allykrupina.github.io/) |
[Code base](https://github.com/allykrupina/allykrupina.github.io)

Please, review.


AMashoshyna:
Great job!

Please consider the following improvements:
- naming: functions names should contain verbs. Exception: words like "handler", "action" etc. which make it clear that the referred item is an action. `initial_cats` is a rather misleading name.
- lines 38-39 and 49-50 are almost identical. Please abstract this logic into a function.
- manipulating DOM nodes is expensive in terms of performance. Try to reduce queries to DOM whenever possible. Example: you make query `document.querySelectorAll(".menu-item")` on every loop iteration. Remember, that the browser needs to traverse the whole DOM to find the elements. Modern browsers are optimized and fast, but still it is a relatively slow operation. Instead you can make the query, save result in a variable and work with it.
- adding event listeners in a loop is ineffective. Please use event delegation instead: add one listener to container element. Also using `element.addEventListener` is more flexible than `onclick`.

allykrupina:
> Great job!
> 
> Please consider the following improvements:
> 
> * naming: functions names should contain verbs. Exception: words like "handler", "action" etc. which make it clear that the referred item is an action. `initial_cats` is a rather misleading name.
> * lines 38-39 and 49-50 are almost identical. Please abstract this logic into a function.
> * manipulating DOM nodes is expensive in terms of performance. Try to reduce queries to DOM whenever possible. Example: you make query `document.querySelectorAll(".menu-item")` on every loop iteration. Remember, that the browser needs to traverse the whole DOM to find the elements. Modern browsers are optimized and fast, but still it is a relatively slow operation. Instead you can make the query, save result in a variable and work with it.
> * adding event listeners in a loop is ineffective. Please use event delegation instead: add one listener to container element. Also using `element.addEventListener` is more flexible than `onclick`.

Hello Anastasiya! 
I considered the improvements so I changes the logic of my code and rewrote it. Also I added a function call with a parameter to the handler. I would like to ask You about the last paragraph of improvements - Why  adding event listeners in a loop is inefficient? If the function is called with a specific parameter..

submissions/allykrupina/js-dom/js/main.js
```diff
@@ -31,22 +31,17 @@ let all_cats = [
...
 ]
 
-function initial_cats() {
-    let handler = function() {
-        let cat_arr = document.querySelectorAll(".menu-item"),
-            index = Array.prototype.indexOf.call(cat_arr, this);
-            cat_title.textContent = all_cats[index].title;
-            cat_img.setAttribute("src", all_cats[index].url);
+function showCats() {
+    drowOneCat = (index) => {
```
AMashoshyna: `const` keyword is missing. Without it it's a global variable.

AMashoshyna:
In this particular case with only six elements you would not see any difference, but event delegation is a useful technique to learn, because handling many events in one place may make the code clearer and more elegant. Also for cases when you have many elements (let's say few thousand cells in a table) adding event listeners to every single one would make a difference in terms of performance.
So I agree that your solution is fine for this case.

submissions/allykrupina/js-dom/js/main.js
allykrupina: Fixed it.Thank you!

AMashoshyna:
LGTM

PR #153, opened 2018-12-25
friends application

IrynaY:
### Friends App

[Demo](https://irynay.github.io/friends-app/) |
[Code base](https://github.com/IrynaY/friends-app)

Please, review.


AMashoshyna:
Very good job! I really like your approach to filters.
Some comments:
- DOM manipulations - better to do in one batch, so:
```
    filteredList.forEach(user => {
      USER_CONTAINER.append(createUserCard(user))
    })
```
Wrap the elements in a [documentFragment](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment) and then append it to container.
- refactoring: lines 77 and 120 are identical. Let's put this code in a function `displayUsersList(list)`.
- also, since logic of getting the data from API is completely separate fron everything else, I would also wrap it in a function `getUsers`
- relying on tag names (example: line 95) is risky, because layout may change. It it better to stick with `data-`-attributes, classes etc.

IrynaY:
@AMashoshyna Thanks for review) I fix all)

IrynaY:
@AMashoshyna  Thanx, done)

AMashoshyna:
LGTM

PR #156, opened 2018-12-26
Add a-tiny-js-world

CuteShaun:
### a-tiny-JS-world

[Demo](https://cuteshaun.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/CuteShaun/a-tiny-JS-world)

Please, review.


submissions/CuteShaun/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,91 @@
...
+const woman = new MainForm("human", 2, 2, "female", "Eva", "Hello, Andrew!", [
+  "Andrew",
+  " Barkl"
+]);
+const dog = new MainForm("animal", 4, 0, "male", "Barkl", "bark", [
+  "Eva",
+  " Andrew"
+]);
+const cat = new Cat("animal", 4, 0, "female", "Masya", "meow", ["Andrew"]);
+const catWoman = new Cat("human", 4, "female", "Eva", ["Andrew", "Barkl"]);
```
OleksiyRudenko: Seems like something got skipped at `catWoman` initialization.

submissions/CuteShaun/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+man.say();
+woman.say();
+dog.say();
+catWoman.say();
+cat.say();
+
+man.dossier();
```
OleksiyRudenko: Please, make so that all inhabitants report themselves in a uniform way, as per tech requirements.

submissions/CuteShaun/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,91 @@
...
+class Cat extends MainForm {
+  constructor(type, legs, hands, gender, name, friendList) {
+    super(type, legs, hands, gender, name, friendList);
+    this.saying = "meow";
+  }
+}
+
+const man = new MainForm("human", 2, 2, "male", "Andrew", "Hello, Eva!", [
+  "Eva",
+  " Masya"
```
OleksiyRudenko: Is space in `" Masya"` a part of name?
Same question re other similar cases across the code.

submissions/CuteShaun/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,99 @@
...
+    this.name = name;
+    this.saying = saying;
+    this.friendList = friendList;
+  }
+
+  say() {
+    print(this.saying);
+  }
+
+  dossier() {
```
OleksiyRudenko: Class should be agnostic of media (`print` is strongly bound to particular media).
Exclusion is classes that deal with media.

Make this method (and `say` as well) so it returns a string.
You would then instruct app to `print(object.dossier())` in the main part of code.

Here is bonus for the implementation:
If you name this method `toString` you will override one of Object magic methods.
By default, e.g. `console.log(somerObj)` logs smth like `[Object object]`. When you override this magic method reference to an object in string context results in whatever `toString` returns.

submissions/CuteShaun/a-tiny-js-world/index.js
```diff
@@ -7,6 +7,21 @@
 
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
+
+//Return string of all property obj values
+function toString(obj) {
+  const valuesArr = Object.values(obj);
```
OleksiyRudenko: Your approach using `allProps` was absolutely correct.
Now you employ `Object` method, which doesn't guarantee properties order (and values therefore as well).
Is there a good reason to change the approach?

OleksiyRudenko: You and MDN are right. The order of props produced by `Object.*` and `for...in` is exactly the same.
But the order itself is not guaranteed. I.e. there is no guarantee that `name` comes before `legs`.

Imagine your project grows. Other contributors extend your classes, add inhabitants etc. That's the case when `Object.*` will fire back.

Please, bring back use of `allProps`. This is also an almost required step to implement and benefit from methods override that you will learn later.

CuteShaun: @OleksiyRudenko, additional tasks and recomendations, both is totally useful and great, thx.

submissions/CuteShaun/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,98 @@
...
+function toString(obj) {
+  return allProps
+    .map(item => {
+      if (Array.isArray(obj[item])) {
+        return obj[item].join(", ");
+      }
+      return `${obj[item]}`;
+    })
+    .join("; ");
+}
```
OleksiyRudenko: Now make `toString` a method of your base class and enjoy how magically `print(man)` works, as `toString` is called on backstage.

PR #157, opened 2018-12-26
add "A Tiny JS World"

SergSenras:
### A Tiny JS World

[Demo](https://sergsenras.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/SergSenras/kottans-frontend/tree/master/task_tiny_js_world)

Please, review.


submissions/SergSenras/tiny-js-world/index.js
```diff
@@ -0,0 +1,80 @@
...
+  
+    Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+    However, please, REFRAIN from improving visuals at least until your code is reviewed
+    so code reviewers might focus on a single file that is index.js.
+  */
+  
+  var family = [man, woman, dog, cat, Catwoman];
+  family.forEach(function(member){
+    var arr = [];
+    for(var key in member){
```
OleksiyRudenko: The order of keys in an object is not guaranteed. Imagine your peer decides to add yet another 100 of objects and defines object properties in different order.
Please, find a way to ensure a very particular order.

submissions/SergSenras/tiny-js-world/index.js
```diff
@@ -0,0 +1,80 @@
...
+      arr.push(member[key]);
+    }
+    print(arr);
```
OleksiyRudenko: Please, ensure that properties are delimited with `;`. Friends, if any, can be delimited with `,`.

submissions/SergSenras/tiny-js-world/index.js
```diff
@@ -0,0 +1,80 @@
...
+const man = {
+    species: 'man',
+    name: 'John',
+    gender: 'male',
+    legs: 2,
+    hands: 2,
+    saying: 'Hi'
+  };
+  const woman = {
+    species: 'woman',
```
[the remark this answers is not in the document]
SergSenras:
```suggestion
const woman = {
    species: 'human',
    name: 'Mary',
```

submissions/SergSenras/tiny-js-world/index.js
SergSenras:
```suggestion
var family = [man, woman, dog, cat, Catwoman];
  function say(obj){
    print('species: ' +obj.species+
          '; name: '  +obj.name+
          '; gender: '+obj.gender+
          '; legs: '  +obj.legs+
          '; hands: ' +obj.hands+
          '; saying: '+obj.saying+
          '; friends: '+obj.friends);
  }
  family.forEach(function(member){
    say(member);
  })
```

submissions/SergSenras/tiny-js-world/index.js
SergSenras:
```suggestion
print('species: '+obj.species+'; name: '+obj.name+'; gender: '+obj.gender+'; legs: '+obj.legs+'; hands: '+obj.hands+'; saying: '+obj.saying+'; friends: '+obj.friends)}


```

submissions/SergSenras/tiny-js-world/index.js
```diff
@@ -0,0 +1,90 @@
...
+  var family = [man, woman, dog, cat, Catwoman];
+  function say(obj){
+    print('species: ' +obj.species+
+          '; name: '  +obj.name+
+          '; gender: '+obj.gender+
+          '; legs: '  +obj.legs+
+          '; hands: ' +obj.hands+
+          '; saying: '+obj.saying+
+          '; friends: '+obj.friends);
+  }
```
OleksiyRudenko: There is a pattern here as all items are delimited with `;`.
How about declaring an array of properties with labels attached and then applying `join`?
The benefit is that it will be better scalable and maintainable. Another developer will just add an element to the array if it is decided that object must have yet another property. Also rearrangement will be easier, and change of delimiter would be done in a single place (argument of `join`).

Please, also follow concern split principle. Current implementation of `say` means that if it is decided to change the media (currently web page and console), we will have to fix the `say` method. Let it just return a well formatted string. This way we also follow single responsibility principle.

SergSenras:
```suggestion
  var family = [man, woman, dog, cat, Catwoman];
  var props = ['species','name','gender','legs','hands','saying','friends'];
  
  family.forEach(member => {
    const data = props.map(prop => {
      return (prop + ': ' + member[prop]);
    })
    print(data.join('; '));
  })
```

PR #159, opened 2018-12-26
Add a-tiny-JS-world

Vitaminvp:
### a-tiny-JS-world

[Demo](https://vitaminvp.github.io/a-tiny-JS-world/) | [Code base](https://github.com/Vitaminvp/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/Vitaminvp/a-tiny-js-world/index.js
+++ b/submissions/Vitaminvp/a-tiny-js-world/index.js
@@ -12,8 +12,8 @@ const man = {
     species: 'human',
     gender: 'male',
     name: 'John',
-    legs: '2',
-    hands: '2',
+    legs: 2,
+    hands: 2,
     saying: 'boys will be boys.',
     friends: []
 };
@@ -21,8 +21,8 @@ const woman = {
     species: 'human',
     gender: 'female',
     name: 'Sarah',
-    legs: '2',
-    hands: '2',
+    legs: 2,
+    hands: 2,
     saying: 'Love saves the world.',
     friends: []
 };
@@ -30,8 +30,8 @@ const dog = {
     species: 'dog',
     gender: 'male',
     name: 'Toby',
-    legs: '4',
-    hands: '0',
+    legs: 4,
+    hands: 0,
     saying: 'woof-woof!',
     friends: []
 };
@@ -39,8 +39,8 @@ const cat = {
     species: 'cat',
     gender: 'female',
     name: 'Maya',
-    legs: '4',
-    hands: '0',
+    legs: 4,
+    hands: 0,
     saying: 'meow',
     friends: []
 };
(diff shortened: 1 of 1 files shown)


submissions/Vitaminvp/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,91 @@
...
+    name: 'Sophie',
+    legs: '2',
+    hands: '2',
+    saying: cat.saying,
+    friends: []
+};
+man.friends = [woman.name, dog.name];
+woman.friends = [man.name, cat.name];
+dog.friends = [man.name, woman.name];
+cat.friends = [woman.name];
```
OleksiyRudenko: Friends are rather other inhabitants than just their names.
You are not required to fix this now, but will be required to fix in a later task.

submissions/Vitaminvp/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,91 @@
...
+catWoman.friends = [];
+
+function printCreature(creature){
+    print(`${creature.species}; ${creature.gender}; <strong>${creature.name}</strong>; ${creature.legs}; ${creature.hands}; <em>${creature.saying}</em>; ${creature.friends.length === 0 ? '' : creature.friends.join(', ')}`);
```
OleksiyRudenko: Please, refactor this method so it returns a string rather than calls `print` (split of concerns principle) so you don't need to rewrite this function when someone decides that different UI method (not `print`) to be applied.

There is also a pattern. Please, form an array of properties (some of those are decorated as per your preference) and than apply `#Array.join`. Why? The latter is better scalable/maintainable if it is decided that object need to have more props or props have to be reordered.

submissions/Vitaminvp/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+const man = {
+    species: 'human',
+    gender: 'male',
+    name: 'John',
+    legs: '2',
+    hands: '2',
```
OleksiyRudenko: Legs and hands are contextually numeric. Is there any good reason to specify those as strings?
Please, explain if there is or fix otherwise.

submissions/Vitaminvp/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,111 @@
...
+			}else if(prop === 'name'){
+				propToPush = `<strong>${creature[prop]}</strong>`;
+			}else if(prop === 'saying'){
+				propToPush = `<em style="text-decoration: underline">${creature[prop]}</em>`;
+			}else {
+				propToPush = creature[prop];
+			}
+			
+            propsArr.push(propToPush);
+        }
```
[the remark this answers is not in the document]
Vitaminvp: I liked last variant, very interesting approach. Thank you.

Vitaminvp:
Please review.

submissions/Vitaminvp/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,94 @@
...
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+
+print(printCreature(man));
+print(printCreature(woman));
+print(printCreature(dog));
+print(printCreature(cat));
+print(printCreature(catWoman));
```
OleksiyRudenko: You could list your creatures as an array and have `.forEach` applied.
Please, read more about `Array` methods, these are very powerful and helpful, also make code better readable.

PR #161, opened 2018-12-26
a Tiny JS World OOP

rrrds:
### a Tiny JS World OOP

[Demo](https://rrrds.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/rrrds/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/rrrds/a-tiny-js-world-oop/index.js
+++ b/submissions/rrrds/a-tiny-js-world-oop/index.js
@@ -7,23 +7,10 @@
 
 // ======== OBJECTS DEFINITIONS ========
 function Creature(props) {
-  this.species = props.species;
-  this.legs = props.legs;
-  this.hands = props.hands;
-  this.name = props.name;
-  this.gender = props.gender;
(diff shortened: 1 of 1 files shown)


submissions/rrrds/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,59 @@
...
+  this.name = props.name;
+  this.gender = props.gender;
+  this.say = props.say;
+}
+Creature.prototype.toString = function() {
+  const values = [];
+
+  for (const prop in this) {
+    if (this.hasOwnProperty(prop)) {
+      values.push(this[prop]);
```
yevhenorlov: This works, however, it introduces unnecessary complexity into code. Should you ever need to make some properties of the Creature private (and inaccessible via `.toString` method), this logic will have to be redone completely.

Why not save props in `this.props` directly, as an object? All you need to do here then is simply list the values of that object as an array and join them.

PR #168, opened 2018-12-26
Add A tiny JS world task

olek557:
### A tiny JS world

[Demo](https://olek557.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/olek557/a-tiny-JS-world/tree/populate-world)

Please, review.

### DIFF
--- a/submissions/olek557/a-tiny-JS-world/index.js
+++ b/submissions/olek557/a-tiny-JS-world/index.js
@@ -1,0 +1,26 @@
+// ======== OBJECTS DEFINITIONS ========
+function Inhabitant(species, name, legs, hands, gender, phrase) {
+   return {
+      species: species,
+      name: name,
+      legs: legs,
+      hands: hands,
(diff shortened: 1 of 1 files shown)


submissions/olek557/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,26 @@
...
+      species: species,
+      name: name,
+      legs: legs,
+      hands: hands,
+      gender: gender,
+      phrase: phrase,
+      // friends: arguments.slice(6),
+      getInfo: function() {
+         return [species, name, legs, hands, gender, phrase].join("; ");
+      }
```
OleksiyRudenko: Well, if we change e.g. name after object construction `.getInfo()` will still report initial value.
You are not required to fix this now. Keep in mind the approach is not really good, and there is no good reason for `getInfo` to be a function.

PR #172, opened 2018-12-27
Add arcade game Frogger

Vitaminvp:
### Arcade game Frogger

[Demo](https://vitaminvp.github.io/kottans-frontend/task_Frogger/) |
[Code base](https://github.com/Vitaminvp/kottans-frontend/tree/master/task_Frogger)

Please, review.

### DIFF
--- a/submissions/Vitaminvp/js-oop-frogger/app.js
+++ b/submissions/Vitaminvp/js-oop-frogger/app.js
@@ -1,0 +1,113 @@
+const cellWidth = 100;
+const cellHeight = 80;
+const enemyWidth = 80;
+const enemyHeight = 60;
+const initialX = 200;
+const initialY = 400;
+const SPRITES = [
+    'images/char-boy.png',
+    'images/char-cat-girl.png',
+    'images/char-horn-girl.png',
+    'images/char-pink-girl.png',
+    'images/char-princess-girl.png'
+];
+function getRandomSpeed() {
+    return Math.round(Math.random() * 600);
+}
+
+// Enemies our player must avoid
+var Enemy = function(x, y, speed) {
+    // Variables applied to each of our instances go here,
+    // we've provided one for you to get started
+    this.x = x;
+    this.y = y;
+    this.speed = speed;
+    // The image/sprite for our enemies, this uses
+    // a helper we've provided to easily load images
+    this.sprite = 'images/enemy-bug.png';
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@Vitaminvp Looks nice! Please, check yourself against the following list:

 * Requirements re **OOP**:
 * Most common mistakes
         
Vitaminvp:
Thanks. I tried to fix code according to remarks mentioned above.

submissions/Vitaminvp/js-oop-frogger/app.js
```diff
@@ -0,0 +1,122 @@
...
+var Player = function(x, y) {
+    this.player = SPRITES[0];//"images/char-boy.png";
+    this.x = x;
+    this.y = y;
+};
+
+Player.prototype.update = function() {};
+
+Player.prototype.render = function() {
+    ctx.drawImage(Resources.get(this.player), this.x, this.y);
```
OleksiyRudenko: Notice how similar are `render` methods from `Player` and `Enemy`. This is a good sign of the fact that both characters classes can benefit from having a base class, so you once you decide to change rendering approach you do this in a single place.
Imagine that you also have gems, health items, different types of enemies etc.

You are not required to fix this though. Just keep in mind.

Vitaminvp:
Thanks. Could you please look again.

submissions/Vitaminvp/js-oop-frogger/app.js
```diff
@@ -0,0 +1,125 @@
...
+    // You should multiply any movement by the dt parameter
+    // which will ensure the game runs at the same speed for
+    // all computers.
+    this.x += this.speed * dt;
+
+    handleCollision(this.player, this);
+
+};
+
+function handleCollision(player, enemy) {
```
OleksiyRudenko: > Please, put collision handling into a dedicated method.

The difference between method and function is that method is a function that belongs to a class.

Please, fix.

OleksiyRudenko:
Please, keep in mind that resolving conversations you have elaborated on is a good way to let reviewer know you did.

Vitaminvp:
How about that?

Vitaminvp:
Thanks :-)

Vitaminvp:
I really appreciate your attention to details. Thanks again.

PR #179, opened 2018-12-28
Memory pair game

denislukianenko:
### Memory Pair Game

[Demo](https://denislukianenko.github.io/memory-pair-game/) |
[Code base](https://github.com/denislukianenko/memory-pair-game)

Please, review.


submissions/denislukianenko/memory-pair-game/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Technologies Pair Game</title>
+    <script src="main.js" defer></script>
```
A-Ostrovnyy: Good practice: tag `<script>` should be before closing tag `<body>`.

submissions/denislukianenko/memory-pair-game/index.html
```diff
@@ -11,5 +11,9 @@
   <body>
     <h1>Technologies Pair Game</h1>
     <div class="field"></div>
+    <div class="modal">
+      <h1>You win!</h1>
```
A-Ostrovnyy: Tag `h1` must be one on the page. You already have one with game title.

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -0,0 +1,93 @@
...
+  font-weight: 800;
+  font-size: 40px;
+}
+
+.field {
+  width: 100%;
+  max-width: 650px;
+  display: grid;
+  grid-gap: 1vw;
+  grid-template-rows: 1fr 1fr 1fr;
```
A-Ostrovnyy: in this case use repeat() function [https://developer.mozilla.org/en-US/docs/Web/CSS/repeat](https://developer.mozilla.org/en-US/docs/Web/CSS/repeat).

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+.card.flipped .both-sides {
+  transform: rotateY(180deg);
+}
+
+.both-sides {
+  height: 100%;
+  width: 100%;
+
```
A-Ostrovnyy: Remove blank lines inside block-rules.

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -0,0 +1,93 @@
...
+  transform-style: preserve-3d;
+
+  position: relative;
+}
+
+.back,
+.front {
+  height: 100%;
+  width: 100%;
+  background-color: rgb(224, 244, 252);
```
A-Ostrovnyy: Try to use one color values.
Below you use HEX and HSL color values.

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -0,0 +1,93 @@
+body {
```
A-Ostrovnyy: Avoid styling by tag add custom styles on class.

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -0,0 +1,93 @@
...
+    Ubuntu, Cantarell, "Open Sans", "Helvetica Neue", sans-serif;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 20px;
+}
+
+*,
+*::before,
+*::after {
```
A-Ostrovnyy: Remove unused rules

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -91,3 +128,48 @@ h1 {
...
+  position: absolute;
+  width: 100vw;
+  height: 100vh;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  display: none;
+
+  background-color: #fffc;
```
A-Ostrovnyy: Is that correct value of color?

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -31,6 +31,43 @@ h1 {
   grid-template-columns: 1fr 1fr 1fr 1fr;
 }
 
+.field > .card:nth-of-type(1) {
+  animation: fadeIn both 600ms 0ms cubic-bezier(0.215, 0.61, 0.355, 1);
```
A-Ostrovnyy: Maybe better way to set `animation` property to all `.field > .card` and reassign only one property `animation-delay` in each `:nth-of-type(n)`.

submissions/denislukianenko/memory-pair-game/style.css
denislukianenko: Got it!

submissions/denislukianenko/memory-pair-game/style.css
denislukianenko: Only * is used. Removed others

submissions/denislukianenko/memory-pair-game/style.css
denislukianenko: Yeah, it actually works like #ffffffcc (with alpha). Raplaced it with hsla, anyways

submissions/denislukianenko/memory-pair-game/style.css
denislukianenko: Fixed!

submissions/denislukianenko/memory-pair-game/main.js
```diff
@@ -0,0 +1,127 @@
...
+  shuffleCards();
+  spreadCards();
+}
+
+function shuffleCards() {
+  technologies.forEach(element => {
+    playArray.push(element);
+    playArray.push(element);
+  });
+  playArray.sort(function() {
```
IgorKurkov: suggest to you move callback to separate function like `randomize` or as u wish because you can use this approach (on another cases) many times - DRY

submissions/denislukianenko/memory-pair-game/main.js
```diff
@@ -0,0 +1,127 @@
...
+  inGuessArray = [];
+  guessedArray = [];
+
+  shuffleCards();
+  spreadCards();
+}
+
+function shuffleCards() {
+  technologies.forEach(element => {
+    playArray.push(element);
```
IgorKurkov: double push - no good but simple. Maybe suggestion make concatenation of technologies array to itself first will be better? You can think about it, for your choice

submissions/denislukianenko/memory-pair-game/main.js
```diff
@@ -0,0 +1,127 @@
...
+    playArray.push(element);
+  });
+  playArray.sort(function() {
+    return 0.5 - Math.random();
+  });
+}
+
+function spreadCards() {
+  let field = document.querySelector(".field");
```
IgorKurkov: here variable from config

submissions/denislukianenko/memory-pair-game/main.js
```diff
@@ -0,0 +1,127 @@
...
+  playArray.forEach(function(el) {
+    let card = document.createElement("div");
+    card.classList.add("card");
+    card.classList.add(el);
+    card.setAttribute("onclick", "handleGuess(this)");
```
IgorKurkov: better approach use event delegation, it means that better you make one eventListener on the container and listen to events on children by `target` property. Here you add a handler to each card. Its not so good but not so bad, but event delegation is must have and you must know this approach and understand how to realize it properly.

submissions/denislukianenko/memory-pair-game/main.js
```diff
@@ -0,0 +1,127 @@
...
+
+    let bothSides = document.createElement("div");
```
IgorKurkov: here you make elements by js on the fly. its okay, but you can make helper function which will create dom element and assign to it some class which you will set as argument like createElement('div', 'both-sides'). but in this case, as I see you try to build card markup, so you can do this by template string and pure HTML inside it. it will be more readable and obvious that you will build and render these HTML cards to the container and it reduce space and rows of code

submissions/denislukianenko/memory-pair-game/main.js
```diff
@@ -0,0 +1,127 @@
...
+      }
+    }
+    render();
+  }
+}
+
+shuffleCards();
+spreadCards();
+
+let modal = document.getElementsByClassName("modal")[0];
```
IgorKurkov: `[0]` looks terrible if you have one modal and it is not an item of a bunch of items. change it to id and set it as a variable to settings obj with `getElementById`

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -15,12 +19,6 @@ body.scroll-blocked {
   overflow: hidden;
 }
 
-*,
-*::before,
-*::after {
-  box-sizing: border-box;
-}
-
 h1 {
```
A-Ostrovnyy: Avoid styling by tag add custom styles on class.

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -139,7 +136,7 @@ h1 {
   justify-content: center;
   align-items: center;
   display: none;
-  background-color: #fffc;
+  background-color: hsla(0, 0%, 100%, 0.8);
   animation: fadeIn both 600ms cubic-bezier(0.215, 0.61, 0.355, 1);
 }
 .modal-shown {
```
A-Ostrovnyy: More readable class name in similar situations will be `.modal-active` or `.modal-isActive`.

submissions/denislukianenko/memory-pair-game/style.css
denislukianenko: Fixed `.scroll-blocked`
I use `.back > img` and `#modal > button` selectors 'cause I'm sure that I want these styles to apply for any image which is a direct child of .back; any button which is a direct child of modal. Avoiding creation of additional classes in HTML seems like a good practice to me in this case, though I understand that with larger and more complicated projects it can cause problems

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -126,24 +126,25 @@ h1 {
   max-height: 70px;
 }
 
-.modal {
+#modal {
```
A-Ostrovnyy: Avoid styling by id.

submissions/denislukianenko/memory-pair-game/style.css
```diff
@@ -126,24 +126,25 @@ h1 {
...
   height: 100vh;
-  display: flex;
   flex-direction: column;
   justify-content: center;
   align-items: center;
-  display: none;
   background-color: hsla(0, 0%, 100%, 0.8);
   animation: fadeIn both 600ms cubic-bezier(0.215, 0.61, 0.355, 1);
 }
+.modal {
```
A-Ostrovnyy: In this case better name `.modal--hide` or `.modal--hidden`. Easier to understand purpose of current class just reading his name.

A-Ostrovnyy:
Good work!

PR #184, opened 2018-12-28
JS-dom

AlexNugda:
### DOM-api task

[Demo](https://alexnugda.github.io/js-dom/) |
[Code base](https://github.com/AlexNugda/frontend-2019-homeworks/tree/dom-api-task/submissions/AlexNugda/js-dom)

Please, review.


OleksiyRudenko:
@AlexNugda Can you please provide a working demo?
It helps to.
Checks other similar PRs for possible solutions on demo publishing.

AMashoshyna:
Hi @AlexNugda!
The demo is still expected, but here are first impressions from the code:
1. Try to rewrite your code so that you don't rely on the order of elements. Examples: use of `document.images[i]` and `element.firstElementChild`. This is not a flexible approach, because the layout may change after redesign and your code will break. Possible solutions: use ids, classes, `data-*` attributes to have reliable and controlled identifiers of DOM elements.
2. To reset `active` class on all elements after click, you can query them directly via document.querySelector('.active').
3. For loop is not optimal choice to find matching element, because after the element is found the loop goes on until end. How about array methods like `Array.prototype.find`?
4. Code style: please make indentations consistent and remove unnecessary empty lines. Also please take a look at the last lines of files: you may notice a small red arrow, which indicates missing newline:

image

AlexNugda:
### DOM-api task
Hi! Thanks for your comments.
I added a link to the demo and corrected the code in accordance with your recommendations.

[Demo](https://alexnugda.github.io/js-dom/) |
[Code base](https://github.com/AlexNugda/frontend-2019-homeworks/tree/dom-api-task/submissions/AlexNugda/js-dom)

Please, review.

submissions/AlexNugda/js-dom/script.js
```diff
@@ -30,32 +30,36 @@ let pages = [ {
...
- 			 h2.textContent = pages[i].articleTitle;
- 			 p1.textContent = pages[i].articleText;
- 			 img.src = pages[i].articleImage;
-			 img2.src = pages[i].articleImage2;
- 		}
- 	}
- } );
\ No newline at end of file
+	let searchPage = (el, i, arr) => {
+		if(e.target.textContent == el.articleTitle)
```
AMashoshyna: Function provided as callback to `Array.prototype.find` should return a Boolean value (`true` if the element matches the search condition, `false` otherwise). Function `searchPage` works as expected, because any object is converted to `true` in Boolean context, and if function returns nothing, it actually return `undefined`, which is converted to `false`, but this is just not how you use Array.prorotype.find. 
This is something related to code semantics. Your solution works (hey, we are in JavaScript land here =) ), but when you see Array.prototype.find method in the code and you need to know what is being searched, you just expect to see a function which answers the question "Is this element ok?"
Suggestion:
```
const searchPage = el => e.target.textContent === el.articleTitle;
```
Also, the functions should be put in const, not let, because you don't want to rewrite the variable by accident. And if you don't use index and array in your code, don't declare variables `i` and `arr`.

AlexNugda:
Thanks for your comments and advice!
I understood and corrected my mistakes.
Please, review.

AMashoshyna:
LGTM

PR #187, opened 2018-12-28
js-pre-oop

OlgaFrontend:
### [Demo](https://olgafrontend.github.io/kottans-frontend/task_js_pre-oop/) |  [Code](https://github.com/OlgaFrontend/kottans-frontend/tree/master/task_js_pre-oop)

Please, review.

### DIFF
--- a/submissions/OlgaDubynka/js-pre-oop/index.js
+++ b/submissions/OlgaDubynka/js-pre-oop/index.js
@@ -30,7 +30,7 @@ const man = new Inhabitant('Rob', 28, 'human', 'male', 2, 2, 0, 0, 'Hi!' );
 // ======== OUTPUT ========
 
 const getStringFromObj = (obj) => {
-    return `<strong>${obj.name}</strong>; age: ${obj.age}; species: ${obj.species}; gender: ${obj.gender}; legs: ${obj.legs}; hands: ${obj.hands}; paws: ${obj.paws}; tail: ${obj.tail}; greeting: ${obj.greeting};`
+    return [`<strong>${obj.name}</strong>`, `age: ${obj.age}`, `species: ${obj.species}`, `gender: ${obj.gender}`, `legs: ${obj.legs}`, `hands: ${obj.hands}`, `paws: ${obj.paws}`, `tail: ${obj.tail}`, `greeting: ${obj.greeting}`].join('; ');
 }
 
 [dog, cat, woman, man].forEach(obj => print(getStringFromObj(obj)));
--- a/submissions/OlgaDubynka/js-pre-oop/index.js
+++ b/submissions/OlgaDubynka/js-pre-oop/index.js
@@ -8,74 +8,32 @@
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
 
-const inhabitants = [
-    {
-        name: 'Richard',
-        age: 5,
-        species: 'dog',
-        gender: 'male',
-        legs: 0,
-        hands: 0,
-        paws: 4,
-        tail: 1,
-        greeting: 'wow!'
-    },
-    {
-        name: 'Rocky',
-        age: 4,
-        species: 'cat',
-        gender: 'female',
-        legs: 0,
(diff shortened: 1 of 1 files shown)


submissions/OlgaDubynka/js-pre-oop/index.js
```diff
@@ -0,0 +1,70 @@
...
+  legs: 2,
+  greeting: 'Hi!'
+}
+
+// ======== OUTPUT ========
+print(dog.name + ';' + dog.age + ';' + dog.gender + ';' + dog.paws + ';' + dog.tail + ';' + dog.greeting + ';');
+print(cat.name + ';' + cat.age + ';' + cat.gender + ';' + cat.paws + ';' + cat.tail + ';' + cat.greeting + ';');
+print(woman.name + ';' + woman.age + ';' + woman.gender + ';' + woman.hands + ';' + woman.legs + ';' + woman.greeting + ';');
+print(man.name + ';' + man.age + ';' + man.gender + ';' + man.hands + ';' + man.legs + ';' + man.greeting + ';');
+/* Use print(message) for output.
```
OleksiyRudenko: There is some repetition above. Please, add a function to form a string from an object.
You may also put your objects into an array and apply `.forEach` to print each one.

OlgaFrontend: Hello, is it ok if I insert the info about objects into small tables, instead of a string? or I should implement exactly in string format?

OleksiyRudenko: @OlgaFrontend Updated code still forms a string just with a different delimiter (`</td><td>`) between chunks of data, plus  some prepending text (`<table><tr><td>`) and appending text in the end (`</td></tr></table>`).
The comment above is still valid. Please, fix accordingly.
Should you require any clarifications, please, let me know.

OleksiyRudenko: @OlgaFrontend Do you need any support on refactoring this part?

To give you a hint on how tables can be formed efficiently and in a scalable manner:

```
return '<table><tr><td>'
  + [
    'cell text',
    aVariable,
    `${aVariable}`,
    `some text: ${aVariable}`,
    ].join('</td><td>')
  + '</td></tr></table>';
```

P.S. Actually building tables in either way is not the best approach, but let's just improve what you've already got.

OleksiyRudenko: @OlgaFrontend Ok, I can see you have reverted code back to plain string.
Is there a good reason not to employ `[...].join(...)`?
My previous comment contains a hint on improvement. This approach will make your code better readable and also scalable.
We write code for the other people, firsthand. And we also want it to be maintainable and scalable.
`Array` methods were also introduced to support us on this.
Please, fix.

submissions/OlgaDubynka/js-pre-oop/index.js
```diff
@@ -30,7 +30,7 @@ const man = new Inhabitant('Rob', 28, 'human', 'male', 2, 2, 0, 0, 'Hi!' );
 // ======== OUTPUT ========
 
 const getStringFromObj = (obj) => {
-    return `<strong>${obj.name}</strong>; age: ${obj.age}; species: ${obj.species}; gender: ${obj.gender}; legs: ${obj.legs}; hands: ${obj.hands}; paws: ${obj.paws}; tail: ${obj.tail}; greeting: ${obj.greeting};`
+    return [`<strong>${obj.name}</strong>`, `age: ${obj.age}`, `species: ${obj.species}`, `gender: ${obj.gender}`, `legs: ${obj.legs}`, `hands: ${obj.hands}`, `paws: ${obj.paws}`, `tail: ${obj.tail}`, `greeting: ${obj.greeting}`].join('; ');
```
OleksiyRudenko: When a list contains many elements or they are quite complex having a single element per line adds to readability.
Example:
```
return [
    'abc',
    'def',
  ].join();
```

PR #188, opened 2018-12-29
object-oriented-js (frogger) - done

Karasb888:
object-oriented-js (frogger)

[Demo](https://karasb888.github.io/frogger-game/) |
[Code base](https://github.com/Karasb888/frogger-game/blob/master/js/app.js)

Please, review.

### DIFF
--- a/submissions/Karasb888/object-oriented-js/app.js
+++ b/submissions/Karasb888/object-oriented-js/app.js
@@ -1,35 +1,43 @@
 var lives = 5;
 var wins = 0;
 
-function setScore(){
-
-  if(lives === 0){
+function setScore() {
+  if (lives === 0) {
     table.innerHTML = 'YOU LOSE';
     alert('YOU LOSE');
-  }else{
-  table.innerHTML = 'Lives: ' + lives + ' | Wins: ' + wins;
+  } else {
+    table.innerHTML = 'Lives: ' + lives + ' | Wins: ' + wins;
   }
-
 };
 
-document.addEventListener('DOMContentLoaded', function(){
-
+document.addEventListener('DOMContentLoaded', function() {
   const table = document.getElementById('table');
   setScore();
-
 });
 
 //MAIN constans
-var GAME_PARAMS = {
+var gameParams = {
   canvas_width: 505,
   canvas_height: 606,
   row_height: 83,
-  coll_width: 101,
-  coll_numb: 6,
+  colums_width: 101,
+  colums_numb: 6,
   player_start_x: 202,
   player_start_y: 398
 };
 
+//coeficient to fix picture size (enemy, player) and game(canvas) object
+SIZING_COEFICIENT = 70;
+//coeficient to fix enemys position in rows
+POSITION_COEFICIENT = 10;
+//Maximum speed of enemy (from 1 to below)
+MAX_SPEED = 6;
+//winning row
+WIN_ROW = 6;
+//starting row
+START_ROW = 1;
+//number of enemy
+ENEMIES = 3;
 //sprites
 const SPRITES = {
   player: 'images/char-boy.png',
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@Karasb888 Looks pretty good at the first glance.

Please, self check yourself against the list below:

 * Requirements re **OOP**:
   * [ ] Separation of Concerns principle is followed
         (e.g. `update` method does only rendering and 
         doesn't **contain** any inline code to check e.g. collisions)
   * [ ] the code is very [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)
 * Most common mistakes
         
Karasb888:
I hurried with OOP, that was rly bad idea to try end this task (and another one) in one night=). After second look I see that I have global mistakes and need to improve my knowledge of OOP=)

Karasb888:
Will fix it as soon as I can=)

OleksiyRudenko:
@Karasb888 This task code is really good. Many things from the list seem to be implemented.
Just go through the code and check against the list to make sure everything is taken into the account.

Karasb888:
So, I do my best =)

submissions/Karasb888/object-oriented-js/app.js
```diff
@@ -0,0 +1,164 @@
...
+    table.innerHTML = 'YOU LOSE';
+    alert('YOU LOSE');
+  }else{
+  table.innerHTML = 'Lives: ' + lives + ' | Wins: ' + wins;
+  }
+
+};
+
+document.addEventListener('DOMContentLoaded', function(){
+
```
OleksiyRudenko: Empty lines right after function signature (and right before closing `}` at function end) do not add to code readability. Please, remove here and across entire code base.

submissions/Karasb888/object-oriented-js/app.js
```diff
@@ -0,0 +1,164 @@
...
+  const table = document.getElementById('table');
+  setScore();
+
+});
+
+//MAIN constans
+var GAME_PARAMS = {
+  canvas_width: 505,
```
OleksiyRudenko: In JS snake case (`abc_def`) is used when entity name is all caps. Please, apply camel case here.
All caps prefix (e..g. `GAME_PARAMS` in `GAME_PARAMS.canvasWidth`) clearly tells we deal with a constant.

submissions/Karasb888/object-oriented-js/app.js
```diff
@@ -0,0 +1,164 @@
...
+  canvas_height: 606,
+  row_height: 83,
+  coll_width: 101,
+  coll_numb: 6,
```
OleksiyRudenko: Pretty ambiguous shortening. Do not shorten words when it save 2 or 3 characters. 
Does `coll` stand for `collocation`, `collection` or `column`?
We write code for other people, firsthand. Some of them will be seeing for the first time.

submissions/Karasb888/object-oriented-js/app.js
```diff
@@ -0,0 +1,164 @@
...
+Enemy.prototype = Object.create(Character.prototype);
+
+Enemy.prototype.start = function() {
+  this.x = -GAME_PARAMS.coll_width;
+  this.speed = getRandomArbitrary(1, 6);
+};
+
+Enemy.prototype.collision = function() {
+  //On collision with player
+  if(this.x  >= player.x - 70 && this.x - 70 <= player.x && this.row === player.row ){
```
OleksiyRudenko: `70` is a magic number. Please, define it as a constant.
Please, find other similar magic values across the code base and define those as constants.

submissions/Karasb888/object-oriented-js/app.js
```diff
@@ -0,0 +1,164 @@
...
+      if(this.x + GAME_PARAMS.coll_width < GAME_PARAMS.canvas_width){
+        this.x += GAME_PARAMS.coll_width;
+      }
+    break;
+
+    case 'up':
+        this.y -= GAME_PARAMS.row_height;
+        this.row++;
+    break;
+
```
OleksiyRudenko: When having `case` fragments properly indented no need to have empty lines anywhere inside `switch`.
```
case 'up':
  this.y -= GAME_PARAMS.row_height;
  this.row++;
  break;
```

Please, install linters and code prettifying plugins to your IDE. They do all dirty job for you formatting code as per conventions.

submissions/Karasb888/object-oriented-js/app.js
```diff
@@ -0,0 +1,164 @@
...
+};
+
+var player = new Player();
+
+var allEnemies = [];
+for (var i = 1; i <= 3; i ++){
+  allEnemies.push( new Enemy(i, player) );
+}
+// This listens for key presses and sends the keys to your
```
OleksiyRudenko: `Array` methods are powerful.
```
var allEnemies = [1, 2, 3].map(row => new Enemy(row, player));
```
You are not required to fix this. Keep the technique in mind.

submissions/Karasb888/object-oriented-js/app.js
OleksiyRudenko: Also `player` here is a global variable.
Please, fix.

submissions/Karasb888/object-oriented-js/app.js
```diff
@@ -0,0 +1,164 @@
...
+};
+
+Enemy.prototype = Object.create(Character.prototype);
+
+Enemy.prototype.start = function() {
+  this.x = -GAME_PARAMS.coll_width;
+  this.speed = getRandomArbitrary(1, 6);
+};
+
+Enemy.prototype.collision = function() {
```
OleksiyRudenko: By convention functions are actions and their names should start with a verb.
`collision` is a noun. Please, refactor.

PR #190, opened 2018-12-29
Dom api task

Humminggoo:
### DOM

[Demo](https://humminggoo.github.io/DOM_menu/) |
[Code base](https://github.com/Humminggoo/DOM_menu)

Please, review.

### DIFF
--- a/submissions/Humminggoo/js-dom/src/index.js
+++ b/submissions/Humminggoo/js-dom/src/index.js
@@ -12,30 +12,30 @@ document.addEventListener("DOMContentLoaded", handleDOM);
 function handleDOM(){
     let menuBtn = document.getElementById('menu-btn'),
         menu = document.getElementById('menu'),
-        menuItems = document.getElementsByClassName('nav-item'),
+        menuItems = document.querySelectorAll('.nav-item'),
+        menuIcon = document.getElementById('menu-btn-icon'),
         contentContainer = document.getElementById('content');
 //for the first load
     showContent(data.projects[0].name);
     menuItems[0].classList.add('nav-item-clicked', 'nav-item-active');
 
 //menu animation
-    menuBtn.addEventListener("click", toggleMenu.bind(null, menu, 'menu-active'));
+    menuBtn.addEventListener("click", () => toggleMenu(menu, menuIcon, 'menu-active'));
 
 //menu items animation
-    menu.addEventListener("click", clickItem, false);
-    menu.addEventListener("mouseenter", highlightItem, true);
-    menu.addEventListener("mouseleave", blurItem, true);
+    menu.addEventListener("click", clickItem);
+    menu.addEventListener("mouseover", toggleItem);
+    menu.addEventListener("mouseout", toggleItem);
 
 //changing open/close menu icons
-    function toggleMenu(menu, classname) {
-        menuBtn.getElementsByTagName('i').item(0).classList.toggle('fa-bars');
-        menuBtn.getElementsByTagName('i').item(0).classList.toggle('fa-times');
+    function toggleMenu(menu, icon, classname) {
+        icon.classList.toggle('fa-bars');
+        icon.classList.toggle('fa-times');
         menu.classList.toggle(classname);
     }
     function clickItem(e) {
-        //apply animation only to menu items
         if(e.target.classList.contains('nav-item')){
-            Array.prototype.forEach.call(menuItems, i => {
+            menuItems.forEach(i => {
                     //if there is already selected item - remove classes from it
                     if(i.classList.contains('nav-item-clicked')){
                         i.classList.remove('nav-item-clicked', 'nav-item-active');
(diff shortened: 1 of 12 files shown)


submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+    let menuBtn = document.getElementById('menu-btn'),
+        menu = document.getElementById('menu'),
+        menuItems = document.getElementsByClassName('nav-item'),
+        contentContainer = document.getElementById('content');
+//for the first load
+    showContent(data.projects[0].name);
+    menuItems[0].classList.add('nav-item-clicked', 'nav-item-active');
+
+//menu animation
+    menuBtn.addEventListener("click", toggleMenu.bind(null, menu, 'menu-active'));
```
zonzujiro: You don't need `bind` here. You can just use arrow function.

submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+
+//menu items animation
+    menu.addEventListener("click", clickItem, false);
+    menu.addEventListener("mouseenter", highlightItem, true);
```
zonzujiro: And what the reason for `capture` here?

submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+        contentContainer = document.getElementById('content');
+//for the first load
+    showContent(data.projects[0].name);
+    menuItems[0].classList.add('nav-item-clicked', 'nav-item-active');
+
+//menu animation
+    menuBtn.addEventListener("click", toggleMenu.bind(null, menu, 'menu-active'));
+
+//menu items animation
+    menu.addEventListener("click", clickItem, false);
```
zonzujiro: `false` is redundant because it is a default value.

submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+    menu.addEventListener("mouseenter", highlightItem, true);
+    menu.addEventListener("mouseleave", blurItem, true);
+
+//changing open/close menu icons
+    function toggleMenu(menu, classname) {
+        menuBtn.getElementsByTagName('i').item(0).classList.toggle('fa-bars');
+        menuBtn.getElementsByTagName('i').item(0).classList.toggle('fa-times');
```
zonzujiro: Do not rely on tagnames or something like this. You can change your design tomorrow and your code will be broken. Use attributes instead.

Also, by using `menuBtn.getElementsByTagName('i').item(0)` you rely on order of your elements. Which is almost as `children[0]`, `parentNode`, `nextSibling`, etc. So, again, in case when you will have changed design - your code will be broken. Which is bad, obviously. 

Use `querySelector` or smth instead.

submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+        menu.classList.toggle(classname);
+    }
+    function clickItem(e) {
+        //apply animation only to menu items
+        if(e.target.classList.contains('nav-item')){
+            Array.prototype.forEach.call(menuItems, i => {
```
zonzujiro: `NodeList` has its own `forEach`

submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+    menu.addEventListener("mouseleave", blurItem, true);
+
+//changing open/close menu icons
+    function toggleMenu(menu, classname) {
+        menuBtn.getElementsByTagName('i').item(0).classList.toggle('fa-bars');
+        menuBtn.getElementsByTagName('i').item(0).classList.toggle('fa-times');
+        menu.classList.toggle(classname);
+    }
+    function clickItem(e) {
+        //apply animation only to menu items
```
zonzujiro: This is a redundant comment. Comments must describe why something is going on. Not what. 

submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+    function highlightItem(e) {
+        if(e.target.classList.contains('nav-item')) {
+            if (!e.target.classList.contains('nav-item-clicked')) {
+                e.target.classList.add('nav-item-active');
+            }
+        }
+    }
+    function blurItem(e) {
+        if(e.target.classList.contains('nav-item')) {
```
zonzujiro: I believe you can combine this in to one `if`. 

Also, maybe you can just remove class name? Without a check?

submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+            if (!e.target.classList.contains('nav-item-clicked')) {
+                e.target.classList.remove('nav-item-active');
+            }
+        }
+    }
+//Dynamic content, obtained from json by menu item name
+    function showContent(name) {
+        let content = data.projects.filter(i => {
```
zonzujiro: `Array#find`?

submissions/Humminggoo/js-dom/src/index.js
```diff
@@ -0,0 +1,101 @@
...
+
+        let nameDiv = document.createElement('div');//project name
+        nameDiv.textContent = content.name;
+        nameDiv.setAttribute('id', 'content-name');
+        nameDiv.classList.add('content-name');
+
+        fragment.appendChild(nameDiv);
+
+        //list of tasks
+        for(let i = 0; i < content.tasks.length; i++){
```
zonzujiro: You should write in the declarative, not imperative way. 

Useful links: 

Humminggoo:
Understood, I won't use Webpack for next projects. I guess, i fixed all, you pointed out. Thanks for detailed explanation.

PR #192, opened 2018-12-29
js-dom

vladk96:
### DOM task

[Demo](https://vladk96.github.io/js-dom/) |
[Code base](https://github.com/vladk96/frontend-2019-homeworks/tree/dom-api-task/submissions/vladk96/js-dom)

Please, review.


submissions/vladk96/js-dom/js/scripts.js
```diff
@@ -0,0 +1,146 @@
...
+
+/*
+    This function search content object in data array
+    and create a new article with this content
+*/
+const changeContent = (navItem) => {
+    clearContent();
+    navItem.classList.add('active');
+    let contentObj = data.find(obj => {
+        if (navItem.textContent.toLowerCase() === obj['name'].toLowerCase()) {
```
IgorKurkov: its okay but in general better way is found obj by id so you can set id's in your `data` container and make a more comfortable search by id, one thing - your titles can be changed:)

submissions/vladk96/js-dom/js/scripts.js
```diff
@@ -0,0 +1,146 @@
...
+            return true;
+        } else {
```
IgorKurkov: also here you can use the ternary operator `return (condition) ? true : false`

vladk96:
Is this pull request approved or I must wait for @yevhenorlov  decision?

submissions/vladk96/js-dom/css/styles.css
```diff
@@ -0,0 +1,113 @@
...
+    background-color: #f3f3f3;
+    font-family: 'Roboto', sans-serif;
+    font-weight: 400;
+}
+
+.menu-button {
+    position: fixed;
+    right: 5%;
+    top: 5%;
+    display: inline-block;
```
yevhenorlov: (see my comment in the HTML section) 
this should be `<button>`, so you don't need to set the display after changing the HTML

submissions/vladk96/js-dom/css/styles.css
```diff
@@ -0,0 +1,113 @@
...
+    height: 100vh;
+    overflow: hidden;
+    width: 0%;
+    box-shadow: 0px 0px 20px 0px #332f2f26;
+    background: white;
+    transition: width .4s linear;
+}
+
+.navigation.open {
+    transition: all .4s linear;
```
yevhenorlov: two things:
1. always explicitly state the properties you are transitioning, otherwise it's a big performance hit. `all` is bad, `width` is good.
2. you don't really need this rule here, it's gonna be inherited from `.navigation` anyway :)

submissions/vladk96/js-dom/css/styles.css
```diff
@@ -0,0 +1,113 @@
...
+}
+
+@media screen and (max-width: 704px) {
+    .navigation.open { width: 200px; }
+}
+
+/* When screen height < 410px,
+	 last nav items do not fit */
+@media screen and (max-height: 410px) {
+    .navigation { overflow: scroll; }
```
yevhenorlov: I don't think you need a specific rule here — just set it to `overflow: scroll` by default for all sizes. 
or maybe I am missing some specific browser case..?

submissions/vladk96/js-dom/css/styles.css
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.menu-close {
+    position: absolute;
+    right: 4%;
+    top: 2%;
+    display: inline-block;
```
yevhenorlov: same as `.menu-button`, see above

submissions/vladk96/js-dom/css/styles.css
```diff
@@ -0,0 +1,113 @@
...
+
+.nav-list {
+    margin: 5rem 0;
+    padding: 0;
+}
+
+.nav-list li {
+    font-size: 1.1rem;
+    padding: 1rem 2rem;
+    list-style: none;
```
yevhenorlov: this property is inherited, it's better to apply it on the parent

submissions/vladk96/js-dom/css/styles.css
```diff
@@ -0,0 +1,113 @@
...
+    cursor: pointer;
+    transition: all .3s ease-in-out;
```
yevhenorlov: see above

submissions/vladk96/js-dom/css/styles.css
vladk96: @yevhenorlov but if I set this rule by default for all sizes the scrollbar does not disappear if all elements fit.
bar

vladk96:
@yevhenorlov I've just finished. Please, review.

submissions/vladk96/js-dom/css/styles.css
yevhenorlov: I see. What browser/OS?
Regardless, I don't seem to be able to reproduce this on Chrome/Firefox on MacOS, so I can't propose a good solution right now. I will merge it as is, maybe someone in chat can help with this issue.

submissions/vladk96/js-dom/index.html
```diff
@@ -0,0 +1,26 @@
...
+        <img src="images/menu.svg" alt="menu">
+    </button>
+    <nav class="navigation">
+        <button class="menu-close">
+            <img src="images/cancel-menu.svg" alt="menu">
+        </button>
+    </nav>
+    <main class="main"></main>
+
+    <!-- my script file -->
```
yevhenorlov: you can remove this comment as well :)

submissions/vladk96/js-dom/css/styles.css
vladk96: I reproduce this on Chrome, Firefox and Edge (Windows 10). Each browser show scrollbar when elements are fit.

submissions/vladk96/js-dom/index.html
```diff
@@ -0,0 +1,26 @@
...
+    <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,700" rel="stylesheet">
+    <link rel="stylesheet" href="css/normalize.css">
+    <link rel="stylesheet" href="css/styles.css">
+</head>
+<body>
+    <button class="menu-button">
+        <img src="images/menu.svg" alt="menu">
+    </button>
+    <nav class="navigation">
+        <button class="menu-close">
```
yevhenorlov: `menu-close` does not seem to work in Firefox now. seems to be a z-index issue, please check

submissions/vladk96/js-dom/index.html
vladk96: @yevhenorlov I've fixed it.

submissions/vladk96/js-dom/index.html
vladk96: Yes, `menu-close` does not work in Firefox, but I do not know how to fix it. I set `z-index` to `menu-close` but it does not help.

vladk96: I added some changes to event handler and `menu-close` works now. Is it good or not?

yevhenorlov: it works, but it's not the best solution - `onNavClick` violates the single responsibility principle (it tries to handle navigation and open/close logic simultaneously, which is confusing for an outsider). 
how about removing `menu-close` logic from `onNavClick` entirely, and simply adding the `onMenuButtonClick` handler to `menu-close`? 
also, maybe rename these functions to reflect _what they do_, not _where they are_ in code.

how about:

```
const handleNavigation = (event) => {
    if (event.target.tagName === 'LI') {
        changeContent(event.target);
    }
}

const toggleMenu = (event) => {
    document.querySelector('.navigation').classList.toggle("open");
}

document.addEventListener('DOMContentLoaded', () => {
    initNavList();
    initContent();
    document.querySelector('.navigation').addEventListener('click', handleNavigation);
    document.querySelector('.menu-button').addEventListener('click', toggleMenu);
    document.querySelector('.menu-close').addEventListener('click', toggleMenu);
});
```

vladk96: I have just finished.

yevhenorlov:
looks good, well done! merging.

PR #197, opened 2018-12-29
add memory pair game task

Aldegid:
### Memory Pair Game

[Demo](https://aldegid.github.io/Memory-pair-game/) |
[Code base](https://github.com/Aldegid/aldegid.github.io/tree/master/Memory-pair-game)

Please, review.


submissions/Aldegid/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,121 @@
...
+let flippedCard = false;
+let lockBoard = false;
+let firstCard;
+let secondCard;
+let showedCards = 0;
+
+function createCards() {
+  images = svgImages;
+  let cardContainer = document.createDocumentFragment();
+  images = images.concat(images); //duplicate array of images
```
zonzujiro: No need in this comment

submissions/Aldegid/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,121 @@
...
+    let card = document.createElement('div');
+    let frontPart = document.createElement('div');
+    let backPart = document.createElement('img');
+    card.classList.add('card');
+    card.dataset.item = item;
+    frontPart.classList.add('front-part');
+    backPart.setAttribute('src', item);
+    backPart.classList.add('back-part');
+    card.appendChild(frontPart);
+    card.appendChild(backPart);
```
zonzujiro: You can use `append` instead

submissions/Aldegid/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,121 @@
...
+  });
+  return cardContainer;
+}
+
+function shuffle(arr) {
+  arr.sort(function () { return 0.5 - Math.random() });
+}
+
+function flipCard(e) {
+  if(!e.target.parentNode.classList.contains('card')){
```
zonzujiro: Don't use constructions like `children[0]`, `parentNode`, `nextSibling`, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. 

You can find out a way how you can perform check on event target, not on the parent.

submissions/Aldegid/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,121 @@
...
+  if (!flippedCard) {
+    flippedCard = true;
+    firstCard = currentCard;
+    return;
+  }
+  secondCard = currentCard;
+  isPaired();
+  setTimeout(() => {
+    showWinMessage();
+  }, 500);
```
zonzujiro: Magic number

submissions/Aldegid/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function isPaired() {
+  let isPair = firstCard.dataset.item === secondCard.dataset.item;
+  isPair ? disableCards() : unflipCards();
```
zonzujiro: Ternary operators usually not used in such way. Better will be if you replace this part with simple if statement.

submissions/Aldegid/memory-pair-game/js/script.js
Aldegid: comments deleted

submissions/Aldegid/memory-pair-game/js/script.js
Aldegid: fixed

submissions/Aldegid/memory-pair-game/js/script.js
Aldegid: like that:
```
function createCards() {
  images = svgImages;
  let cardContainer = document.createDocumentFragment();
  images = images.concat(images);
  shuffle(images);
  images.forEach(function (item) {
    let card = document.createElement('div');
    card.classList.add('card');
    card.dataset.item = item;
    card.innerHTML = `<div class="front-part"></div>
                      <img src="${item}" class="back-part">`
    cardContainer.append(card);
  });
  return cardContainer;
}
```

zonzujiro: Kind of.

Or like `card.append(frontPart, backPart)`

submissions/Aldegid/memory-pair-game/js/script.js
zonzujiro: It's in your [code](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad) ;)

Aldegid: thx, accapted)

Aldegid:
I implemented required changed.  Pls review.

PR #204, opened 2018-12-30
Please, review files for DOM task

dnzln:
### DOM

[Demo](https://dnzln.github.io/dom/) |
[Code base](https://github.com/dnzln/dom)

Please, review.


submissions/dnzln/js-dom/style.css
```diff
@@ -0,0 +1,96 @@
+body {
```
A-Ostrovnyy: Avoid styling by tag add custom styles on class.

submissions/dnzln/js-dom/style.css
```diff
@@ -0,0 +1,96 @@
...
+    width: auto; 
+    padding: 0 5px;
+    font-family: Georgia;
+    font-size: 16px;
+    align-content: flex-start;
+}
+
+header {
```
A-Ostrovnyy: In `html` you added class `.main-header` but not use him.

submissions/dnzln/js-dom/style.css
```diff
@@ -0,0 +1,96 @@
...
+
+.menu-item {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+    text-align: center;
+}
+
+li {
+    line-height: 35px;
```
A-Ostrovnyy: Better use relative value of `line-height`. The value of `line-height` should be calculated from the font size. Because when you change `font-size` in media rules you need to change `line-height` too. If you use relative value you change only one property.
Example: `font-size: 10px;` and you need line-height value - 12px so you write `line-height: 1.2;`.

submissions/dnzln/js-dom/style.css
```diff
@@ -0,0 +1,96 @@
...
+li:hover {
+    color: #fff;
+    cursor: pointer;
+}
+
+footer {
+    text-align: center;
+    width: 100%;
+    margin-top: 15px;
+    background-color: #000000;
```
A-Ostrovnyy: Try to use one style for colors value.
Here you use long color value 000000 but below in the code you use short color value #fff.

submissions/dnzln/js-dom/style.css
```diff
@@ -0,0 +1,96 @@
...
+        margin-left: 2%;
+        padding: 0 2% 0 2%;;
+        margin-top: 15px;
+        background-image: linear-gradient(to bottom, #ff0037, #ff2a2a);
+    }
+}
+
+
+@media screen and (min-width: 900px) {
+    body {
```
A-Ostrovnyy: This is bad practice(change width of `body`): add block-wrapper/container for this rules.

dnzln:
Thanks for the review! I corrected mentioned errors in the code.

submissions/dnzln/js-dom/style.css
```diff
@@ -36,48 +39,54 @@ h2 {
...
 
-    aside {
+    .menu {
         width: 25%;
         padding-top: 15px;
         margin-top: 15px;
     }
 
-    main {
+    #main-content {
```
A-Ostrovnyy: Avoid styling by id. Separate responsibilities: `class` for styles `id` for scripts.

submissions/dnzln/js-dom/style.css
```diff
@@ -1,31 +1,34 @@
...
-header {
+.main-header {
     margin: 0px;
     padding: 10px 0;
     text-align: center;
-    background-color: rgb(218, 218, 218);
+    background-color: #d8d8d8;
+}
+
+h2 {
```
A-Ostrovnyy: Avoid styling by tag add custom styles on class.

submissions/dnzln/js-dom/style.css
```diff
@@ -1,31 +1,34 @@
...
+    font-family: Georgia, 'Times New Roman', Times, serif;
+    font-size: 16px;
+}
+
+.wrapper {
     width: auto; 
     padding: 0 5px;
-    font-family: Georgia;
-    font-size: 16px;
     align-content: flex-start;
```
A-Ostrovnyy: This property aligned content in flex container. `.wrapper` not have `display: flex;` so add this property(`align-content`) when you set `.wrapper` `display: flex;`

submissions/dnzln/js-dom/code.js
```diff
@@ -28,8 +28,7 @@ const content = [
 
 
 var container = document.createElement('section');
```
A-Ostrovnyy: Avoid using tags for wrapper without class name.

dnzln:
Done!

submissions/dnzln/js-dom/style.css
```diff
@@ -36,48 +39,54 @@ h2 {
     text-align: center;
 }
 
-li {
-    line-height: 35px;
+.menu-item li {
```
A-Ostrovnyy: Eventually you adding style to tag `li`.
Hint: `<ul class="menu-list">` `<li class="menu-item">`

dnzln:
Thought it's ok to style perhaps h1-h6 tags with tag selector, or cases like lists items. But now I will surely remember to use classes for all tags) Thanks for help!

A-Ostrovnyy:
Good work!

PR #206, opened 2018-12-30
oop tiny js world

yulyasystem:
### post oop tiny js world
[Demo](https://yulyasystem.github.io/task-oop-tiny-js-world/) |
[Code base](https://github.com/yulyasystem/task-oop-tiny-js-world/blob/master/index.js)

Please, review.


submissions/yulyasystem/post-oop-world/index.js
```diff
@@ -0,0 +1,50 @@
+class Entity {
+    constructor(species, name, gender, legs) {
+        this.species = species;
+        this.name = name;
+        this.gender = gender;
+        this.legs = legs;
+
+    }
+    saying() {
```
OleksiyRudenko: By convention function names denote actions and should start with a verb. Please, rename.
If you name this method `toString` there will be a bonus learning for you.

submissions/yulyasystem/post-oop-world/index.js
```diff
@@ -0,0 +1,50 @@
+class Entity {
+    constructor(species, name, gender, legs) {
+        this.species = species;
+        this.name = name;
+        this.gender = gender;
+        this.legs = legs;
+
```
OleksiyRudenko: As per tech spec every inhabitant is expected to have its own saying, individual for humans.
Please, fix here and across other classes as appropriate.

submissions/yulyasystem/post-oop-world/index.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+class Human extends Entity {
+    constructor(name, gender, hands) {
+        super('human', name, gender, 2);
+        this.hands = 2;
+    }
+    saying() {
+        return super.saying() + `I have ${this.hands} hands and ${this.legs} legs. It is no suprise`;
+    }
```
OleksiyRudenko: You have employed `super` extended functionality. That's great.
The benefit of `super.*` is that sub classes do not need to know anything about props defined in their base class. And to be OOP compliant also **should not** know.
So let super class to tell us anything about `legs` and let `Human` to supplement info received with anything re `hands`.
No problem if hands come after anything else. Being OOP compliant is much more important learning.

You have also overloaded `saying`, which means that if you rename a method in base class, methods in child classes should be renamed too.

submissions/yulyasystem/post-oop-world/index.js
```diff
@@ -0,0 +1,50 @@
...
+        super(species, name, gender);
+
+    }
+}
+
+const man = new Human('George', 'male');
+const woman = new Human('Alisa', 'female');
+const cat = new Animal('cat', 'Velik', 'male');
+const dog = new Animal('dog', 'Hugo', 'male');
+const catWoman = new CatWoman('cat-woman', 'Hinata', 'female');
```
OleksiyRudenko: See how you have a class defining human species. And species of dogs and cats didn't deserve for their own, which is ~discrimination~ inconsistency in class hierarchy: `Entity` gets extended by classes of incomparable type - a species and a more general entity.
Please, fix.

submissions/yulyasystem/post-oop-world/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+print(man.saying());
+print(woman.saying());
+print(cat.saying());
+print(dog.saying());
+print(catWoman.saying());
```
OleksiyRudenko: There is a repetition above, which can be easily avoided and optimization will also make code less verbose and more scalability friendly.
Hint: `[....].forEach()`

A bonus learning: once you rename `saying` as `toString` you do not need to specify the method name in string context. `toString` is a magic method (white wizardry) and is called implicitly in string context. So `print(dog)` would do.

OleksiyRudenko:
@yulyasystem Please, also remove all files except `index.js`, which is sufficient for the review.
Thanks.

submissions/yulyasystem/post-oop-world/index.js
yulyasystem: cool HNY magic, I like it, thanks :heart_eyes:

submissions/yulyasystem/post-oop-world/index.js
yulyasystem: Am I right with my solution?

PR #210, opened 2018-12-31
Add oop version of Tiny JS World

Humminggoo:
### OOP exercise

[Demo](https://humminggoo.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Humminggoo/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/Humminggoo/index.js
+++ b/submissions/Humminggoo/index.js
@@ -1,0 +1,97 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+function Being(name, gender, species, hands, legs){
+    this.name = name;
+    this.species = species;
+    this.gender = gender;
+    this.hands = hands;
+    this.legs = legs;
+    this.phrase = this.say();
+}
+Being.prototype = {
+    say: function () {
+        return 'Hello!';
+    }
+};
+function Dog(name, gender, species='dog', hands=0, legs=4){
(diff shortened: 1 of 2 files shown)


AMashoshyna:
One of the learning goals of this exercise is to handle the situation where subclass constructor (**Human** in our case) has more arguments than base class (**Being**). That’s why it is important that only humans have hands. To justify this exercise, let’s imagine the situation when you design your classes and start with **Being** as base class and **Cat** and **Dog** as subclasses. So, at this point you have no reason to include hands. Human class appears later with new client’s requirement and now you need to add ‘hands’ as constructor parameter. Of course, you can add it to **Being** and make it optional or add some default value for creatures who don’t have hands, but then all of the creatures will have this redundant property. Try to find a more elegant solution.

Humminggoo:
Added new classes Human and Animal, made some fixes to existing ones. Will that suffice?

AMashoshyna:
@Humminggoo Nice job! But now for beings without hands the code outputs `undefined`. Let's make the output personalized for every species to avoid this. Hint: if you redefine `toString` method, you can print the object directly, without helper function.

Humminggoo:
Done. Added toString to Being as you suggested.

AMashoshyna:
Well done! Your PR is good to merge.
As further improvement here is a trick you can do with "toString". Since `print` function expects a string, and if something is not a string, the "toString" method is called by the JavaScript engine "under the hood" to provide type coercion, here is what you can do:
```javascript
[dog, cat, woman, man, catwoman].forEach(i => print(i));  // i.toString called
```
Details about "toString" [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString#Description)
Or even better, since "print" expects one argument, and "forEach" provides array item as first argument to callback, you can write like so:
```javascript
[dog, cat, woman, man, catwoman].forEach(print);
```
Cool, right?

PR #216, opened 2018-12-31
update app.js (HNY)

AlinaLadybug:
### Object-Oriented JavaScript

[Demo](https://alinaladybug.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/AlinaLadybug/frontend-nanodegree-arcade-game/tree/master/js)

Please, review.

### DIFF
--- a/submissions/AlinaLadybug/object-oriented/app.js
+++ b/submissions/AlinaLadybug/object-oriented/app.js
@@ -1,18 +1,45 @@
-const CELL = { width: 101, height: 83 };
-const CANVAS = { width: 505, height: 606 };
-const ENEMY = { speed: 100, startPosition: -50, width: 80, height: 60 };
-const PLAYER_LOCATION = { x_axis: 200, y_axis: 400 };
+const CELL = {
+    width: 101,
+    height: 83
+};
+const CANVAS = {
+    width: 505,
+    height: 606,
+    offsetY: 31
+};
+const ENEMY = {
+    speed: 100,
+    startPosition: -50,
+    width: 80,
+    height: 60,
+    sprite: 'images/enemy-bug.png'
+};
+const PLAYER = {
+    xAxis: 200,
+    yAxis: 400,
+    sprite: 'images/char-boy.png'
+};
 
-const Creature = function (x, y) {
+const Creature = function (x, y, sprite) {
     this.x = x;
     this.y = y;
+    this.sprite = sprite;
 }
+
+
+Creature.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+}
+
 // Enemies our player must avoid
-const Enemy = function (x, y, speed) {
-    Creature.call(this, x, y);
+const Enemy = function (x, y, speed, player) {
+    Creature.call(this, x, y, sprite = ENEMY.sprite);
     this.speed = speed;
-    this.sprite = 'images/enemy-bug.png';
+    this.player = player;
 };
+
+
+Enemy.prototype = Object.create(Creature.prototype);
 
 Enemy.prototype.update = function (dt) {
     this.x += this.speed * dt;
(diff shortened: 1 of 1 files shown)


submissions/AlinaLadybug/object-oriented/app.js
```diff
@@ -0,0 +1,87 @@
+const CELL = { width: 101, height: 83 };
+const CANVAS = { width: 505, height: 606 };
+const ENEMY = { speed: 100, startPosition: -50, width: 80, height: 60 };
+const PLAYER_LOCATION = { x_axis: 200, y_axis: 400 };
```
OleksiyRudenko: Please, format the literals as per conventions, one line per property+value. This take some extra lines of code but also adds to readability.

submissions/AlinaLadybug/object-oriented/app.js
```diff
@@ -0,0 +1,87 @@
...
+
+const Creature = function (x, y) {
+    this.x = x;
+    this.y = y;
```
OleksiyRudenko: You have introduced a base class to hold shared properties `x` and `y`.
Please, note that `sprite` is also a shared property and `render()` is a shared method. Please, move those to `Creature`.

submissions/AlinaLadybug/object-oriented/app.js
OleksiyRudenko: Per convention underscore is used to delimit parts of all caps identifiers like `PLAYER_LOCATION`.
When any identifiers aren't all caps, use camelCase. e.g. `xAxis`, like `startPosition`.

submissions/AlinaLadybug/object-oriented/app.js
```diff
@@ -0,0 +1,87 @@
...
+    if (this.x > CANVAS.width) {
+        this.x = ENEMY.startPosition;
+        this.speed = Math.floor(Math.random() * 200) + ENEMY.speed;
+    }
+
+    if (player.x > this.x - ENEMY.width && player.x < this.x + ENEMY.width &&
+        player.y > this.y - ENEMY.height && player.y < this.y + ENEMY.height) {
+        player.x = PLAYER_LOCATION.x_axis;
+        player.y = PLAYER_LOCATION.y_axis;
+    }
```
OleksiyRudenko: This block seem to handle collisions. As this is an important feature it is worth being moved to its own dedicated method to follow the Separation of Concerns principle.

Also `player` is a global variable. Classes should be agnostic of the outer world. The only globals allowed are constants and library functions.

Please, find a way for enemies to know the player explicitly.

submissions/AlinaLadybug/object-oriented/app.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+let allEnemies = [];
+let player = new Player(PLAYER_LOCATION.x_axis, PLAYER_LOCATION.y_axis);
+let enemiesLocation = [52, 132, 212];
+enemiesLocation.forEach(elem => {
+    let enemy = new Enemy(0, elem, ENEMY.speed);
+    allEnemies.push(enemy);
+})
```
OleksiyRudenko: Consider the following example:
```
let allEnemies = [52, 132, 212]
   .map(locationY => new Enemy(0, locationY, ENEMY.speed);
```
Which saves you 66% of lines used before and also geds rid of an unnecessary variable.

Of course `new Enemy(...)` will be supplemented with an extra argument, which is a reference to the player.

Beside that `[52, 132, 212]` are also magic numbers worth being constants. BTW, aren't those calculable from `CELL.height`?
`[1,2,3].map(row => row * CELL.height - CELL.yOffset)...`
where `CELL.yOffset` equals `-31` or so.

AlinaLadybug:
@OleksiyRudenko  Thanks for your comments!)

PR #217, opened 2018-12-31
Happy new year!

nazmariam:
### Object oriented javascript

[Demo](https://nazmariam.github.io/frogger/index.html) |
[Code base](https://github.com/nazmariam/nazmariam.github.io/blob/master/frogger/index.html)

Please, review.


submissions/nazmariam/oop-js/js/app.js
```diff
@@ -0,0 +1,109 @@
...
+
+};
+Player.prototype.toStart = function () {
+    this.x = start.x;
+    this.y = start.y;
+};
+Player.prototype.handleInput = function(code){
+    switch (code) {
+        case 'left':
+            this.x-=100;
```
OleksiyRudenko: This is a magic number. Worth being defined as a constant with a descriptive name. Please, define a const for it and other magic numbers across the code.
You can group them in object literals just like you did with `field` and `start`.

submissions/nazmariam/oop-js/js/app.js
```diff
@@ -0,0 +1,109 @@
+const start = {
+    x: 202,
+    y: 390
+};
+const field = {
+    w: 505,
+    h: 606
+};
```
OleksiyRudenko: There are two major conventions as to const names.
Please, follow the `ALL_CAPS_WITH_UNDERLINES` convention for now.
Fields inside the object const can still be regular `camelCase`.

submissions/nazmariam/oop-js/js/app.js
```diff
@@ -0,0 +1,109 @@
...
+            if (this.y < 0) {
+                win++;
+                console.log(win);
+                alert("Happy New Year! You win "+win+" times!");
+                this.toStart();
+            }
+            break;
+    }
+};
+function checkCollision(enemy) {
```
OleksiyRudenko: This function can effectively be a method of `Enemy`.
You will have to refactor it then.
Neither class should refer to global variables like `player` as classes should be agnostic of the outer world. The only exception is global constants and library methods.
So, when you make this fn an `Enemy` method make sure enemies receive reference to the player explicitly, store it in their props and refer to the player via a prop in `.checkCollision`

submissions/nazmariam/oop-js/js/app.js
```diff
@@ -0,0 +1,109 @@
...
+let allEnemies = [
+    new Enemy(-100, 55, 100),
+    new Enemy(-400, 135, 70),
+    new Enemy(0, 135, 35),
+    new Enemy(0, 220, 50),
+    new Enemy(-300, 220, 70)
+];
+let player = new Player();
+document.addEventListener('keyup', function(e) {
+    var allowedKeys = {
```
OleksiyRudenko: Semantically this a complex const.

submissions/nazmariam/oop-js/js/app.js
nazmariam: Did it. Please, check.

submissions/nazmariam/oop-js/js/app.js
nazmariam: done

submissions/nazmariam/oop-js/js/app.js
nazmariam: done

submissions/nazmariam/oop-js/js/app.js
OleksiyRudenko: ... which means it is better declared as `const` )

submissions/nazmariam/oop-js/js/app.js
OleksiyRudenko: > This is a magic number. Worth being defined as a constant with a descriptive name. Please, define a const for it **and other magic numbers across the code**.
> You can group them in object literals just like you did with field and start.

submissions/nazmariam/oop-js/js/app.js
```diff
@@ -0,0 +1,114 @@
...
+};
+let win = 0;
+let Enemy = function(x,y,speed) {
+    this.sprite = 'images/enemy-bug.png';
+    this.x = x;
+    this.y = y;
+    this.speed = speed;
+};
+Enemy.prototype.update = function(dt,gamer) {
+    this.x += dt*this.speed;
```
OleksiyRudenko: Does your code work?

nazmariam: it works, cause I put player as gamer in this function in engine.js

nazmariam: Fixed without changing engine.

submissions/nazmariam/oop-js/js/app.js
```diff
@@ -0,0 +1,112 @@
...
+    if (
+       this.gamer.x < this.x + STEP.x/2 &&
+        this.gamer.x + STEP.x/2 > this.x &&
+        this.gamer.y < this.y + STEP.y/4 &&
+        this.gamer.y + STEP.y/4 > this.y
+    ) {
+        return true;
+    } else {
+        return false;
+    }
```
OleksiyRudenko: The fragment above can be optimized.
Expression in `if` condition returns `true` or `false` itself, so you can just return it.
E.g.
```
return this.gamer.x < this.x + STEP.x / 2 &&
  this.gamer.x + STEP.x/2 > this.x &&
  whateverElseLogicalTests
```

submissions/nazmariam/oop-js/js/app.js
```diff
@@ -0,0 +1,112 @@
...
+        this.gamer.x = START.x;
+        this.gamer.y = START.y;
+    }
+};
+Enemy.prototype.render = function() {
+    if(this.x >= FIELD.w)
+        this.x = -STEP.x;
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+Enemy.prototype.checkCollision = function(){
```
OleksiyRudenko: By conventions functions that return a boolean start with a special verb, which is normally `is`.
Imagine you're asking a question "Is there a collision?" and a response might be "True, there is". Other modal verbs are also in use (`should`, `can`, `does` etc).

You are not required to fix this. Just keep in mind.
**Please, do not spend time on refactoring this**

submissions/nazmariam/oop-js/js/app.js
```diff
@@ -0,0 +1,112 @@
...
+
+    if (this.checkCollision()){
+        alert("You loose");
+        this.gamer.x = START.x;
+        this.gamer.y = START.y;
+    }
+};
+Enemy.prototype.render = function() {
+    if(this.x >= FIELD.w)
+        this.x = -STEP.x;
```
OleksiyRudenko: This method is dedicated to drawing. You have different method to manage location. So location boundaries tests is more appropriate to have there.

Separation of Concerns principle helps to write more consistent code.

Fix this in your own game repo. No need to update the PR.

PR #219, opened 2019-01-01
Add friends app

Vitaminvp:
### Friends App

[Demo](https://vitaminvp.github.io/kottans-frontend/task-friendApp/) |
[Code base](https://github.com/Vitaminvp/kottans-frontend/tree/master/task-friendApp)

Please, review.


submissions/Vitaminvp/friend-app/js/app.js
```diff
@@ -0,0 +1,187 @@
...
+
+const state = {
+    CardsList: [],
+    filteredList: [],
+    sortedList: []
+};
+
+const SORT = {
+    all:  () => renderCards(state.sortedList = [...state.filteredList]),
+    male: () => renderCards(state.sortedList = [...state.filteredList].filter((card) => card.gender.toLowerCase() === 'male')),
```
AMashoshyna: Why do you need to spread the array here?

Vitaminvp: I wanted to prevent coping link instead of an array, or it is not necessary for object properties?

Vitaminvp:
Please, review.

submissions/Vitaminvp/friend-app/js/app.js
AMashoshyna: It is very nice that you have thought of that. Some comments:
- **Array.prototype.filter** creates a new array and does not modify the initial one, so you can skip array copying with **filter**, and also **map**, **reduce** and some more array methods.
- as fas a s I can see, renderCards does not modify array either. Instead it uses copy of the array created by **Array.prototype.slice**

One more question: do you intend to make an assignment here?
```javascript
all:  () => renderCards(state.sortedList = [...state.filteredList])
```

I think you do it in line 170 before calling SORT.all:
```javascript
state.sortedList.length >= 0 ? null : state.sortedList = [...state.filteredList];
```
I may be missing something, so don't hesitate to correct me.
BTW, I would prefer to see a simple if-statement in the line above. Ternary expressions are good when you use both **if** and **else** branches.

AMashoshyna:
Nice job! Let's remove unnecessary array copying and you are done!

submissions/Vitaminvp/friend-app/js/app.js
Vitaminvp: Line 170 works if only state.sortedList doesn't exist or zero length.
all:  () => renderCards(state.sortedList = [...state.filteredList])
This line is responsible for the initial state of filtering by gender. If I change it, I can’t render both a man and a woman again. I tried.

Everything else - done.

Vitaminvp:
All - done  :)

PR #222, opened 2019-01-01
Please, review A Tiny JS World task

dnzln:
### A Tiny JS World

[Demo](https://dnzln.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/dnzln/a-tiny-JS-world/blob/gh-pages/index.js)

Please, review.


submissions/dnzln/tiny-js-world/index.js
```diff
@@ -0,0 +1,62 @@
...
+    legs: '2',
+    hands: '2',
+    name: 'Alishia',
+    gender: 'female',
+    says: catsTalks
+}
+
+// ======== OUTPUT ========
+print(man.species + ';' + man.legs + ';' + man.hands + ';' + man.name + ';' + man.gender + ';' + man.says);
+print(woman.species + ';' + woman.legs + ';' + woman.hands + ';' + woman.name + ';' + woman.gender + ';' + woman.says);
```
OleksiyRudenko: Please, define a function that will form a string from an object.
The string formation you used has certain pattern. Will you try to employ `Array.join()`?

submissions/dnzln/tiny-js-world/index.js
```diff
@@ -0,0 +1,62 @@
...
+print(dog.species + ';' + dog.legs + ';' + dog.hands + ';' + dog.name + ';' + dog.gender + ';' + dog.says);
+print(cat.species + ';' + cat.legs + ';' + cat.hands + ';' + cat.name + ';' + cat.gender + ';' + cat.says);
+print(catWoman.species + ';' + catWoman.legs + ';' + catWoman.hands + ';' + catWoman.name + ';' + catWoman.gender + ';' + catWoman.says);
```
OleksiyRudenko: Please, ensure  every code file has an empty line at the end. You may tune your IDE to do this dirty job for you.

image

OleksiyRudenko: The series of `print` instructions are also a repetition pattern. How about to list all object in an array and apply `Array.forEach()`?
Example:
```
[variable1, variable2, whateverElseVar].forEach(item => someFunction(item));
```
This will save you many lines and also makes code easier to scale. Imagine, you need to print tens of inhabitants.

OleksiyRudenko:
> To be honest, I don't get how to use Array.join() in this case, with the objects. So I did it in another way. Seems like I miss something, but can you please tell me, how to do it with that method?

Example:
`[a.aString, a.bString].join("; ")`
Does the above explain?

dnzln:
Yes, it's pretty clear now. Thanks! Done.

PR #227, opened 2019-01-02
A tiny JS world

ivarshavets:
### A Tiny JS World

[Demo](https://ivarshavets.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/ivarshavets/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/ivarshavets/a_tiny_js_world/index.js
+++ b/submissions/ivarshavets/a_tiny_js_world/index.js
@@ -1,0 +1,35 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+
+   Code repository: https://github.com/ivarshavets/a-tiny-JS-world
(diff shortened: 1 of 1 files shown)


submissions/ivarshavets/a_tiny_js_world/index.js
```diff
@@ -0,0 +1,35 @@
...
+    this.gender = gender
+    this.legs = legs
+    this.hands = hands
+    this.speak = speak
+    this.friends = friends
+  }
+
+  getInfo(){
+    return Object.values(this).join("; ")
+  }
```
AMashoshyna: Please don't traverse `this` directly. This may have unexpected effects, for example list inherited properties. In this case the safest way to control listed properties is to call them one by one explicitly.

ivarshavets: Thanks! Changed.

PR #234, opened 2019-01-02
Task 8. Object Oriented JS

babayK0:
### Task 8. Object Oriented JS

[Demo](https://babayk0.github.io/Frogger/) |
[Code base](https://github.com/babayK0/babayK0.github.io/blob/master/Frogger/js/app.js)

Please, review.


submissions/babayK0/Object-Oriented-JS/Frogger/js/app.js
```diff
@@ -0,0 +1,99 @@
...
+                this.y = PLAYER_Y_START;
+            }, 1000);
+        }
+    }
+}
+// Now instantiate your objects.
+// Place all enemy objects in an array called allEnemies
+let allEnemies = [];
+ENEMY_Y_POSITION.forEach(function(yPosition) {
+    enemy = new Enemy(ENEMY_START_POSITION, yPosition);
```
AMashoshyna: Here you create a global variable and redefine it on every function iteration. A cleaner approach would be to create local variable with let/const.

submissions/babayK0/Object-Oriented-JS/Frogger/js/app.js
```diff
@@ -0,0 +1,99 @@
...
+// This listens for key presses and sends the keys to your
+// Player.handleInput() method. You don't need to modify this.
+document.addEventListener('keyup', function(e) {
+    var allowedKeys = {
+        37: 'left',
+        38: 'up',
+        39: 'right',
+        40: 'down'
+    };
+
```
AMashoshyna: If the key is not in `allowedKeys`, `player.handleInput` will be called with `undefined`. It does not cause any harm, but still it would be cleaner to check for undefined before calling the `player.handleInput` method.

babayK0: @AMashoshyna but what should we do if the value is "undefined"?. i.e. when we checked "key" and it is not in "allowedKeys" .Can we just continue or we must do ''console.log" or something else like alert sentence?

submissions/babayK0/Object-Oriented-JS/Frogger/js/app.js
babayK0: @AMashoshyna I understand that I am redefine "ENEMY_START_POSITION", or not?I am somewhat confused. Can you explain again. May i do something like this?
```suggestion
    enemy = new Enemy(500, yPosition);
```

submissions/babayK0/Object-Oriented-JS/Frogger/js/app.js
AMashoshyna: Well, this would be a question to a UX-designer (what is the appropriate way to notify the user that he/she pushed wrong key). But in this case I think that not calling the function is enough.

babayK0: @AMashoshyna ok,i will write function but she will not be called ( )

submissions/babayK0/Object-Oriented-JS/Frogger/js/app.js
babayK0: @AMashoshyna oh :crying_cat_face: : i'm inattentive

babayK0:
@AMashoshyna that's all ? Next time i will be more attentive :smile_cat: :

AMashoshyna:
Yeah, that's all =) LGTM

PR #235, opened 2019-01-02
Task 9. OOP exercise

babayK0:
### Task 9. OOP exercise

[Demo](https://babayk0.github.io/a-tiny-JS-world/) | [Code base](https://github.com/babayK0/a-tiny-JS-world/blob/master/index.js)

Please, review.


babayK0:
@AMashoshyna Hello! I think I did everything as you requested. If something is unsuccessful please tell me, I will try to fix it.

AMashoshyna:
Almost perfect! You solved the issue with different sets of properties, but this is not exactly the OOP way. Try to think towards inheritance  and method overriding: what if sub-classes (Human, Animal etc.) implement their own `info` getter?

babayK0:
@AMashoshyna I initially did it (Human and Animal had own "info").But there are a lot same text(I think it will break DRY pattern ).And my decision was to wrote only one "info" with some ternary operators. I will still rewrite but try to do it later.

babayK0:
@AMashoshyna is it ok now? :smile_cat:

AMashoshyna:
Nice job!

PR #236, opened 2019-01-02
Add advanced-OOP task

webdevagent:
### OOP Exercise

[Demo](https://webdevagent.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/webdevagent/frontend-2019-homeworks/blob/advanced-OOP-task/submissions/webdevagent/advanced-oop/index.js)

Please, review.

### DIFF
--- a/submissions/webdevagent/advanced-oop/index.js
+++ b/submissions/webdevagent/advanced-oop/index.js
@@ -2,43 +2,50 @@
 /*Resident is a parent class and all future classes
 will inherit from this class*/
 class Resident {
-  constructor(species, name, gender, legs, hands, voice, friends) {
+  constructor(species, name, gender, voice, friends) {
     this.species = species;
     this.name = name;
     this.gender = gender;
-    this.legs = legs;
-    this.hands = hands;
     this.voice = voice;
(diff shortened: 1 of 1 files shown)


AMashoshyna:
Hi @webdevagent! You did well, but one of the main goals of this exercise is to implement class hierarchy where some classes have more properties than others, for example: only humans have hands, other creatures do not. Those additional properties effect class methods (showResidentInfo must be implemented differently for different sets of properties).
Your solution works, but it misses the point. Although you implement  inheritance by extending Resident class, your sub-classes don't change anything in parent class behavior. Try to find a different approach and use your new OOP knowledge. Key points - inheritance (parent class and sub-classes), composition, method overloading.

webdevagent:
Thank's for the review @AMashoshyna  and Happy New Year!) I have checked my code and I absolutely agree with your suggestion. Yeah, my old versions of a new class defining don't have any sense because they don't change anything. So I defined a new method in parent class : setResidentData(), and overladed it in child classes, so they have some individual properties.

AMashoshyna:
Well done!

PR #237, opened 2019-01-03
add files for dom_task

AnnaGrynchuk:
###task_js_dom

Please, review.

### DIFF
--- a/submissions/AnnGrynchuk/js_dom/index.html
+++ b/submissions/AnnGrynchuk/js_dom/index.html
@@ -1,0 +1,42 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <title>Paradise</title>
+    <link rel="stylesheet" type="text/css" href="style.css">
+  </head>
+  <body>
+    <header >
+      <div >
+         <h1 class="header_name">
+            Most Beautiful Places In The World 
+        </h1>
+       </div>
+    </header>
+    <main>
+      <section >
+          <nav class="side_menu">
+              <ul  class="side_menu_list">
+                <li class="side_menu_item">
+                    Moraine Lake
+                </li>
+                <li class="side_menu_item">
+                    Geiranger Fjord
+                </li>
+                <li class="side_menu_item">
+                    Bora Bora
+                </li>
+                <li class="side_menu_item">
+                    Santorini
+                </li>
+              </ul>
+            
+              <div  class="description">
+              </div>
+          </nav>
+      </section>
+    </main>
+    <script src="main.js"></script>
+  </body>
+ </html>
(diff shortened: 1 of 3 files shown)


submissions/AnnGrynchuk/js_dom/style.css
```diff
@@ -0,0 +1,103 @@
...
+      font-size:28px;
+      text-align:center;
+    }
+
+  .side_menu {
+    display: flex;
+    height: 100%;
+    width: 100%;
+    margin-left: auto;
+    margin-right: auto;
```
yevhenorlov: `margin: 0 auto;` is better, but it's probably not needed at all, containers are 100% wide anyway :)

submissions/AnnGrynchuk/js_dom/main.js
```diff
@@ -0,0 +1,27 @@
...
+    {title: 'Santorini',
+      text: 'Located on top of a cliff with a spectacular view of the Palea volcano, Nea Kameni, and the island of Thirassia, Oia is the most popular and arguably the most beautiful of all the picturesque villages of the beautiful Greek island of Santorini. Only about 11 km from Fira, on the north of the island, Oia will charm you with its traditional stone houses lining the narrow streets, breathtaking blue-domed churches, and sunbaked verandas. <br><br> While the village has its share of taverns, souvenir shops, and cafes, Oia is more quiet and laid-back than busy Fira and most people enjoy its quaint beauty by slowly exploring its narrow streets. Stroll through the village’s small port of Ammoudi by descending 300 steps down the cliff, or enjoy colorful galleries showcasing art from the many artists who fell in love with the village and made it their home. Oia, Santorini is considered by many one of the prettiest places in the world. '}
+    ];
+
+let sideMenu = document.querySelector('.side_menu_list');
+let place = document.querySelectorAll('.side_menu_item');
+let description = document.querySelector('.description');
```
yevhenorlov: you're not reassigning these variables, so it's better to use `const`.

submissions/AnnGrynchuk/js_dom/main.js
```diff
@@ -0,0 +1,27 @@
...
+
+description.innerHTML = infoArray[0].text ;
+
+sideMenu.addEventListener('click', function(e) {
+  place.forEach((item, i) =>{
+    if(place[i]  === e.target){
```
yevhenorlov: you're already passing the `item` parameter, so you can
```javascript
if(item === e.target)
```

submissions/AnnGrynchuk/js_dom/index.html
```diff
@@ -0,0 +1,42 @@
...
+                    Santorini
+                </li>
+              </ul>
+            
+              <div  class="description">
+              </div>
+          </nav>
+      </section>
+    </main>
+    <script src="main.js"></script>
```
yevhenorlov: `<script>` in the end of the `<body>` does the job in this case, but it's not always enough - should you ever need more complex logic in your app, this may introduce bugs the cause of which may not be obvious.
Consider wrapping your javascript code into a function (named `init`, for example), and then calling it when the HTML is parsed by the browser like this:
```javascript
document.addEventListener('DOMContentLoaded', init)
```
This way you can safely put `<script>` anywhere in the html file, it will always wait for the DOM to load before trying to select elements. Also, this way you're not cluttering the global scope of the program, everything is neatly contained within the `init` function scope.

PR #238, opened 2019-01-03
Object Oriented JS

OlgaFrontend:
### [Demo](https://olgafrontend.github.io/kottans-frontend/frogger-game/) |  [Code](https://github.com/OlgaFrontend/kottans-frontend/tree/master/frogger-game)

Please, review.


submissions/OlgaDubynka/js-oop-frogger/js/app.js
```diff
@@ -0,0 +1,95 @@
...
+    
+    }
+}
+
+const allEnemies = [];
+
+const enemyLocation = [60, 145, 225];
+
+enemyLocation.forEach(function (locationY) {
+    enemy = new Enemy(0, locationY, 200);
```
AMashoshyna: Here you create a global variable. It is cleaner to create a local variable with let/const

submissions/OlgaDubynka/js-oop-frogger/js/app.js
```diff
@@ -0,0 +1,95 @@
...
+const player = new Player(202, 405);
+
+document.addEventListener('keyup', function(e) {
+    var allowedKeys = {
+        37: 'left',
+        38: 'up',
+        39: 'right',
+        40: 'down'
+    };
+
```
AMashoshyna: If the key is not in `allowedKeys`, `player.handleInput` will be called with `undefined`. It does not cause any harm, but still it would be cleaner to check for undefined before calling the `player.handleInput` method.

submissions/OlgaDubynka/js-oop-frogger/js/app.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    render () {
+        ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+    }
+
+    update (dt) {
+        this.x += this.speed * dt;
+    
+        if (this.x > 505) {
```
AMashoshyna: please avoid so-called "magic numbers" in code: put the numbers into properly named constants.

AMashoshyna:
Great job!
Let's fix some issues. Additionally to my comments to code, please note a bug in UI:
- I can move the player outside the filed if I click right arrow multiple times. Add a check that the player does not cross game field borders maybe?

PR #239, opened 2019-01-03
add inebitans for tiny JS world

AnnaGrynchuk:

### js-pre-oop

[Code base](https://github.com/AnnaGrynchuk/frontend-2019-homeworks/blob/pre-oop-task/submissions/AnnGrynchuk/js_pre_oop/index.js)

Please, review.

### DIFF
--- a/submissions/AnnGrynchuk/js_pre_oop/index.js
+++ b/submissions/AnnGrynchuk/js_pre_oop/index.js
@@ -7,56 +7,40 @@
 
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
-const dog = {
-    species: 'dog',
+const inhebitants = [ 
+    {species: 'dog',
     name: 'Toby',
     gender: 'male',
     legs: 4,
     hands: 0,
-    saying: 'woof-woof!'
-  };
+    saying: 'woof-woof!'},
 
-const cat = {
-    species: 'cat',
+   {species: 'cat',
     name: 'Persik',
     gender: 'male',
     legs: 4,
     hands: 0,
-    saying: 'meawww-meaww!'
-}
+    saying: 'meawww-meaww!'},
 
-const woman = {
-    species: 'human',
+   {species: 'human',
     name: 'Poly',
     gender: 'female',
     legs: 2,
     hands: 2,
-    saying: 'I am hungry!'
-}
+    saying: 'I am hungry!'},
 
-const man = {
-    species: 'human',
+   {species: 'human',
     name: 'Alex',
     gender: 'male',
     legs: 2,
     hands: 2,
-    saying: 'Lets go to bar!'
-}
+    saying: 'Lets go to bar!'}
+];
 
-print("In a tiny JS world lives " + dog.species + ". Its name is <strong>" +  dog.name  + "</strong>. Its gender is " + dog.gender + ". It has strong " + 
-  dog.legs + " legs and " + dog.hands + " hands. When it wants to play it says " + dog.saying );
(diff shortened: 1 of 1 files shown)


submissions/AnnGrynchuk/js_pre_oop/index.js
```diff
@@ -0,0 +1,63 @@
...
+  dog.legs + " legs and " + dog.hands + " hands. When it wants to play it says " + dog.saying );
+
+print("Wow! We see fluffy " + cat.species + " in our JS world. Its name is <strong>" +  cat.name  + "</strong>. Its gender is " + cat.gender + ". It has long " + 
+  cat.legs + " legs and " + cat.hands + " hands. At 4a.m it always says " + cat.saying );
+
+print("Suddenly beautiful " + woman.species + " appears. Her name is <strong>" +   woman.name  + "</strong>. Her gender is " +  woman.gender + ". She has long " + 
+woman.legs + " legs and thin " +  woman.hands + " hands. She said to her friend " +  woman.saying );
+
+print("Her friend <strong> " + man.name  + " </strong> wasn't surprised, he knows her well. He has strong  " +  man.hands + " hands and  " +  man.legs + " legs so he easily picked up her and said " +  man.saying );
+
```
OleksiyRudenko: Too much repetition. Define a function that takes an object and returns a formatted string. Call it within `print`.

You can also benefit from using `Array` methods like `forEach`.

Example:
`[var1, var2, var3].forEach(item => print(someFunc(item));`

This will make your code shorter and better readable.

submissions/AnnGrynchuk/js_pre_oop/index.js
```diff
@@ -0,0 +1,51 @@
...
+   {species: 'human',
+    name: 'Alex',
+    gender: 'male',
+    legs: 2,
+    hands: 2,
+    saying: 'Lets go to bar!'}
+];
+
+function printInhebitants(item){
+    return Object.values(item).join('-');
```
OleksiyRudenko: Please, note that order of properties in an object is not guaranteed.
So it is more consistent to list properties explicitly in a desired order.
E.g. `[obj.a, obj.b, obj.c].join(.......)`

submissions/AnnGrynchuk/js_pre_oop/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+   {species: 'human',
+    name: 'Alex',
+    gender: 'male',
+    legs: 2,
+    hands: 2,
+    saying: 'Lets go to bar!'}
+];
+
+function printInhebitants(item){
```
OleksiyRudenko: Technically, this function doesn't print inhabitants but rather produces a formatted string.
Name like `convertInhabitantToString` probably not perfect but more precise.

We write code for other people, firsthand. So function name forms certain expectation.

submissions/AnnGrynchuk/js_pre_oop/index.js
```diff
@@ -0,0 +1,51 @@
...
+    return Object.values(item).join('-');
+ }
+
+inhebitants.forEach(item => print(printInhebitants(item)));
+
+inhebitants[2].saying ="I'm hungry too!";
+const newMan = Object.assign({},inhebitants[2]);
+print(newMan.saying);
```
OleksiyRudenko: Seems like you wanted to have a newMan with a different saying.
What you did is changed saying of an existing inhabitant first and copied it into a new instance. You might probably want first copy and then change the saying.

PR #241, opened 2019-01-03
Object Oriented JS

ivarshavets:
### Object Oriented JS - Frogger game

[Demo](https://ivarshavets.github.io/kottans-frontend/frogger/index.html) |
[Code base](https://github.com/ivarshavets/kottans-frontend/blob/master/frogger/js/app.js)

Please, review.


submissions/ivarshavets/js-oop/app.js
```diff
@@ -0,0 +1,110 @@
...
+    this.x = ENEMY.startPosition;
+  }
+
+  if (player.x < this.x + ENEMY.width &&
+    player.x + ENEMY.width > this.x &&
+    player.y < this.y + ENEMY.height &&
+    ENEMY.height + player.y > this.y) {
+    player.reset();
+  };
+};
```
OleksiyRudenko: Separation of Concerns principle should be followed (e.g. `update` method does only rendering and doesn't contain any inline code to check e.g. collisions)

OleksiyRudenko: Classes should not refer to any global variables, like global variable `player`, which is an instance of Player class (referring to global constants and globals provided by the gaming platform like Resources is OK); Hint: pass `Player` instance as an argument to every enemy

submissions/ivarshavets/js-oop/app.js
```diff
@@ -0,0 +1,110 @@
...
+    default:
+      break;
+  }
+}
+
+const player = new Player(PLAYER.startX, PLAYER.startY);
+const allEnemies = [
+  new Enemy(-CELL_WIDTH, 63, 100),
+  new Enemy(-CELL_WIDTH, 145, 50),
+  new Enemy(-CELL_WIDTH, 230, 75)
```
OleksiyRudenko: Numbers above are mysterious.
All numbers like block dimensions, initial locations should be defined as constants.

ivarshavets:
@OleksiyRudenko Thanks! Fixed.

PR #242, opened 2019-01-03
add friends-app task

allykrupina:
### Friends App

[Demo](https://allykrupina.github.io/friends-app/) |
[Code base](https://github.com/allykrupina/allykrupina.github.io/tree/master/friends-app)

Please, review.


submissions/allykrupina/friends-app/index.html
```diff
@@ -0,0 +1,66 @@
...
+                    <input type="radio" id="age1" name="age" value="increase">
+                    <label for="age1">
+                        <i class="fa fa-sort-numeric-asc" aria-hidden="true"></i>
+                    </label>
+                    <input type="radio" id="age2" name="age" value="decrease">
+                    <label for="age2">
+                        <i class="fa fa-sort-numeric-desc" aria-hidden="true"></i>
+                    </label>
+                    <h4>Sort by name</h4>
+                    <input type="radio" id="age3" name="age" value="abc">
```
zonzujiro: It is not `age`, its a name and looks like it is not working.

submissions/allykrupina/friends-app/main.js
```diff
@@ -0,0 +1,116 @@
...
+      SOCIAL = ['instagram', 'facebook', 'linkedin', 'twitter', 'dribbble'];
+let initialArray = [],
+    newArray = [],
+    social = SOCIAL.map((item)=> {
+        return `<a target="_blank" href="https://www.${item}.com"><i class="fa fa-${item}"></i></a>`
+    })
+
+let createItem = item => {
+    initialArray.push(item);
+    newArray = initialArray;
```
zonzujiro: You will do this assigning on every iteration of `Array#map`, which is redundant. It can be done only once.
Also, if you want to make a copy of an array, you can use `Array#slice`

submissions/allykrupina/friends-app/main.js
```diff
@@ -0,0 +1,116 @@
...
+}
+let renderFriendsBlock = friendsArray => {
+    friendsArray.map(createItem)
+    ageFilter.addEventListener('change', ({target}) => filterAge(target.value, newArray));
+    genderFilter.addEventListener('change', ({target}) => filterGender(target.value, newArray));
+    nameFilterInput.addEventListener('keyup', ({target}) => filterName(target.value, initialArray));
```
zonzujiro: Will `keyup` work if I paste value inside `input` with a mouse? 🤔

submissions/allykrupina/friends-app/main.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+let filterAge = (value, array) => {
+    array.sort((a, b) => {
+        if(value === 'increase'){
+            return a.dob.age - b.dob.age;
+        } else if (value === 'decrease') {
+            return b.dob.age - a.dob.age;
+        } else if (value === 'abc') {
+            return a.name.first > b.name.first;
```
zonzujiro: Sorting of an array of strings is [trickier](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort). You should take into action case of a letter and return `1`, `-1` or `0` and not only truthy or falsie values.

submissions/allykrupina/friends-app/main.js
```diff
@@ -0,0 +1,116 @@
...
+        } else {
+            return a.name.first < b.name.first;
+        }
+    });
+    filterRender(array);
+}
+let filterName = (value, array) => {
+    newArray = array.filter(item => item.name.first.indexOf(value) + 1 || item.name.last.indexOf(value) + 1);
```
zonzujiro: `String#includes`

submissions/allykrupina/friends-app/main.js
```diff
@@ -0,0 +1,116 @@
...
+    friendsArray.map(createItem)
+    ageFilter.addEventListener('change', ({target}) => filterAge(target.value, newArray));
+    genderFilter.addEventListener('change', ({target}) => filterGender(target.value, newArray));
+    nameFilterInput.addEventListener('keyup', ({target}) => filterName(target.value, initialArray));
+    resetFilter.addEventListener('click', ({target}) => filterReset(initialArray));
+    new FriendsContainer(friends, newArray);
+};
+
+let filterRender = (array) => {
+    friends.innerHTML = '';
```
zonzujiro: Pls, make more obvious assigning of the variable, because right now:
1. you get it from the global scope
2. in general, create a variable in HTML file

Use `getElementById` instead and create separate variable.
And this I mean for every such element.

submissions/allykrupina/friends-app/main.js
```diff
@@ -0,0 +1,116 @@
...
+            <p class="friend-name">${this.item.name.first} ${this.item.name.last}</p>
+            <p><span class="friend-gender">${this.item.gender}, </span><span class="friend-age">${this.item.dob.age}</span></p>
+            <div class="friend-social">${social.join('')}</div>
+            <p class="friend-detail">
+                I am special agent ${this.item.login.username}. I am from ${this.item.location.city}. My password is ${this.item.location.street} ${this.item.location.state} ${this.item.location.postcode} ${this.item.location.coordinates.longitude} ${this.item.location.coordinates.latitude}<br><br>
+                My email ${this.item.email}.
+            </p>
+        </div>`;
+        block.innerHTML = content;
+        friends.appendChild(block);
```
zonzujiro: Adding elements to DOM from a loop is a bad practice. A browser will run reflow and repaint for every element in the loop. Instead, you can: 

1. Create some wrapper, add your items to the wrapper and then add it to DOM. It will be one operation. 
2. Clone current container. Add items to a container and then replace your old container with a new one. But be aware of event listeners.

submissions/allykrupina/friends-app/main.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+fetch(FRIENDS_API_URL)
+.then(response => response.json())
```
zonzujiro: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/

submissions/allykrupina/friends-app/main.js
```diff
@@ -0,0 +1,116 @@
...
+.then(data => {
+    renderFriendsBlock(data.results);
+})
+.catch(error => {
+    result.innerHTML = error;
+    result.style.display = 'block';
```
zonzujiro: You can use your function `error` instead, right? 🤔

submissions/allykrupina/friends-app/main.js
zonzujiro: You should imply `handleErrors` and add it to your code. Because of reasons from article ;)

allykrupina:
@zonzujiro hope I fixed all changes requested)

submissions/allykrupina/friends-app/index.html
allykrupina: @zonzujiro  
fixed it

PR #247, opened 2019-01-03
OOP exercise

dnzln:
### OOP exercise

[Demo](https://dnzln.github.io/oop-world/) |
[Code base](https://github.com/dnzln/oop-world/blob/master/index.js)

Please, review.

### DIFF
--- a/submissions/dnzln/oop-world/index.js
+++ b/submissions/dnzln/oop-world/index.js
@@ -8,14 +8,15 @@
 // ======== OBJECTS DEFINITIONS ========
 
 class Mammal {
-    constructor(species, name, gender, say) {
+    constructor(species, name, gender, say, legs) {
         this.species = species;
         this.name = name;
         this.gender = gender;
         this.say = say;
+        this.legs = legs;
     }
 
-    setProperties() {
+    formString() {
         return [
             '<strong>Species: </strong>',
             this.species + '\t\t',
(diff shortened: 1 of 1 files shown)


submissions/dnzln/oop-world/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+population.forEach(
+    function(elem) {
+        let arrStr = [];
+        for (var key in elem) {
+            arrStr.push('<strong>'+key+'</strong>'+ ': ');
+            arrStr.push(elem[key] + '\t\t');
+        }
+        print(arrStr.join(''));
+    }
```
OleksiyRudenko: Outer world shouldn't know how a class is organized inside.
Add a method to the base class that will return a formatted string.

Child classes' methods can overload one from parents, which means a child class may define a method with the same name as parent class have.
You can call a method from parent class (`super.methodName()`), supplement it with extra info unique to a child class and return aggregated result.

submissions/dnzln/oop-world/index.js
```diff
@@ -0,0 +1,52 @@
...
+var woman = new Human('Eve', 'female', 'Thug life!');
+var dog = new Animal('dog', 'Snoop', 'male', 'Woof!');
+var cat = new Animal('cat', 'Smelly', 'female', 'Moor!');
+
+var population = [man, woman, dog, cat];
+
+population.forEach(
+    function(elem) {
+        let arrStr = [];
+        for (var key in elem) {
```
OleksiyRudenko: The order of properties in an object is not guaranteed. Please, list properties explicitly.

A hint:
`[var1, var2, var3, obj.a, obj.b]`
`Array` methods can be applied to such array.

dnzln:
Done.

submissions/dnzln/oop-world/index.js
```diff
@@ -0,0 +1,71 @@
...
+        );
+        print(properties.join(''));
+    }
+}
+
+var man = new Human('human', 'Adam', 'male', 'What\'s up?').printOut();
+var woman = new Human('human', 'Eve', 'female', 'Thug life!').printOut();
+var dog = new Animal('dog', 'Snoop', 'male', 'Woof!').printOut();
+var cat = new Animal('cat', 'Smelly', 'female', 'Moor!').printOut();
+
```
OleksiyRudenko: `man === woman === dog === cat === undefined` since `.printOut()` returns nothing (`indefined` implictly).

To give these variables some existential sense separate inhabitants creation from reporting.

You may also benefit from e.g. `[varA, varB].forEach(....);`

dnzln:
I added new tails property to Animal class, because there is no more need in that class in another case. Is it OK now?

OleksiyRudenko:
> I added new tails property to Animal class, because there is no more need in that class in another case. Is it OK now?

You've get the point perfectly right - there is no point in having a class that doesn't add any specificity or uniqueness. You decision to add a property to save the class is pretty OK.

PR #253, opened 2019-01-04
Add memory-pair-game

CuteShaun:
### Memory pair game

[Demo](https://cuteshaun.github.io/Pair_game/) |
[Code base](https://github.com/CuteShaun/Pair_game)

Please, review.


submissions/CuteShaun/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,100 @@
...
+    "img/stan.png",
+    "img/bill.png",
+    "img/soos.png"
+  ];
+
+  const container = document.getElementById("parent");
+  const heading = document.getElementById("heading");
+  let html;
+
+  //////////////////////Function data handlers
```
zonzujiro: This is a redundant comment. Comments must describe why something is going on. Not what. 

submissions/CuteShaun/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,100 @@
...
+    activeElements.forEach(item => {
+      item.classList.add("hidden");
+    });
+  }
+
+  function compareImgSrc(activeElements) {
+    return (
+      activeElements.length > 1 &&
+      activeElements[0].children[0].children[1].children[0].src ===
+        activeElements[1].children[0].children[1].children[0].src
```
zonzujiro: Don't use constructions like `children[0]`, `parentNode`, `nextSibling`, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use `querySelector` or smth instead.

submissions/CuteShaun/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,100 @@
...
+    );
+  }
+
+  function getActiveElems(activeElements) {
+    activeElements = [...document.querySelectorAll(".active")];
+    return activeElements;
```
zonzujiro: Why are you pass `activeElements` here as an argument? 🤔

submissions/CuteShaun/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  //////////////////////Event listener with cards logic
+  container.addEventListener("click", function(event) {
+    let classList = event.target.classList;
+    getActiveElems();
+
+    if (getActiveElems().length > 1) {
```
zonzujiro: You already have active elements in your variable. Why do you need this call?

submissions/CuteShaun/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,100 @@
...
+
+    if (classList.contains("flip-box")) {
+      classList.toggle("active");
+      getActiveElems();
+    }
+
+    if (compareImgSrc(getActiveElems())) {
+      addHidden(getActiveElems());
+      let quantityHidden = document.querySelectorAll(".hidden");
+      console.log(quantityHidden);
```
zonzujiro: Be aware that you should remove all `console.log` in your code before creating a PR.

submissions/CuteShaun/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,100 @@
...
+
+      //When all cards hidden, change main-heading color
+      quantityHidden.length === 12
+        ? heading.classList.add("main-heading--active")
+        : console.log("нихуяшечки");
```
zonzujiro: Ok...
First, not all developers will like such humor. Second, if something goes wrong, not all clients will like such humor 😉 

Pls, remove it.

submissions/CuteShaun/memory-pair-game/js/script.js
CuteShaun: Thanks for this comment, I keep in mind this :)

submissions/CuteShaun/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,93 @@
...
+
+  function addHidden(activeElements) {
+    activeElements.forEach(item => {
+      item.classList.add("hidden");
+    });
+  }
+
+  function compareImgSrc(imgs) {
+    imgs = [...document.querySelectorAll(".active .flip-box__img")];
+    if (imgs.length === 2 && imgs[0].src === imgs[1].src) {
```
zonzujiro: Just return check. You don't need `return true` and `if` here

submissions/CuteShaun/memory-pair-game/js/script.js
```diff
@@ -0,0 +1,93 @@
...
+        <div class="flip-box__back">
+          <img class="flip-box__img"src="${item}" alt="cartoon-character" />
+        </div>
+      </div>
+    </div>`;
+      })
+      .join(" ");
+
+    container.innerHTML = html;
+    return;
```
zonzujiro: Redundant return

PR #254, opened 2019-01-04
Memory – Pair Game

dnzln:
### Memory – Pair Game

[Demo](https://dnzln.github.io/memory-card/) |
[Code base](https://github.com/dnzln/memory-card/tree/master)

Please, review.


zonzujiro:
kapture 2019-01-04 at 10 31 37

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
+var arr = [];
+var arrayImg = [1,2,3,4,5,6,1,2,3,4,5,6];
+arrayImg = arrayImg.sort(function() { return 0.5 - Math.random() });  // перемешиваем массив
+
+arrayImg.forEach(                       // создаем на странице div с классом back и в него вкладываем
```
zonzujiro: Comment only in english

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
+var arr = [];
+var arrayImg = [1,2,3,4,5,6,1,2,3,4,5,6];
+arrayImg = arrayImg.sort(function() { return 0.5 - Math.random() });  // перемешиваем массив
```
zonzujiro: This is a redundant comment. Comments must describe why something is going on. Not what. 

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
...
+
+arrayImg.forEach(                       // создаем на странице div с классом back и в него вкладываем
+    function(elem, ind, arr) {          // скрытую картинку. Блок и картинка добавляется для каждой карточки
+        let underPicture = document.createElement('img');
+        underPicture.classList.add('under-img');
+        underPicture.setAttribute('src', 'img/'+elem+'.png');
+        let conteiner = document.getElementById(ind);
```
zonzujiro: "container" :)

You can use spellcheckers for editors. As an example here is mine for VSC - https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

For Chrome you can use Grammarly

Also, here you will search for the same element on each loop iteration. Better save it to a variable outside the loop.

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
+var arr = [];
+var arrayImg = [1,2,3,4,5,6,1,2,3,4,5,6];
+arrayImg = arrayImg.sort(function() { return 0.5 - Math.random() });  // перемешиваем массив
+
+arrayImg.forEach(                       // создаем на странице div с классом back и в него вкладываем
+    function(elem, ind, arr) {          // скрытую картинку. Блок и картинка добавляется для каждой карточки
```
zonzujiro: Unused variable `arr`

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
...
+        let underPicture = document.createElement('img');
+        underPicture.classList.add('under-img');
+        underPicture.setAttribute('src', 'img/'+elem+'.png');
```
zonzujiro: You can use template literals for such things

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
...
+        let conteiner = document.getElementById(ind);
+        let underConteiner = document.createElement('div');
+        underConteiner.classList.add('back');
+        underConteiner.appendChild(underPicture);
+        conteiner.appendChild(underConteiner);
```
zonzujiro: Adding elements to DOM from a loop is a bad practice. A browser will run reflow and repaint for every element in the loop. Instead, you can: 

1. Create some wrapper, add your items to the wrapper and then add it to DOM. It will be one operation. 
2. Clone current container. Add items to a container and then replace your old container with a new one. But be aware of event listeners.

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+);
+
+for(let i = 0; i <= 11; i++) {
+    document.getElementById(i).addEventListener('click', function () {main(i)});  // каждой карте - слушатель!
```
zonzujiro: You should use event delegation where it is possible. Adding of listener to each element – is not very effective. 

Take a look on nice answer about this technique: https://stackoverflow.com/questions/1687296/what-is-dom-event-delegation.

Also, prefer array methods to loops.

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+function main(id) {     // каждый клик обрабатывается по сценарию ниже, вызываются соответ. функции: 
+    arr.push(id);       // открытие или закрытие карточек, скрытие отыграных карточек.
```
zonzujiro: Dnt use shrtd vrbls nms 😉 

Also, `arr` to general name. Find something more specific, like `openedCards`, as an example.

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
...
+    document.getElementById(i).addEventListener('click', function () {main(i)});  // каждой карте - слушатель!
+}
+
+function main(id) {     // каждый клик обрабатывается по сценарию ниже, вызываются соответ. функции: 
+    arr.push(id);       // открытие или закрытие карточек, скрытие отыграных карточек. 
+    openImg(id);
+    if(arr.length == 1) return;
+    if(arr.length == 2) {
+        if(arrayImg[arr[0]] == arrayImg[arr[1]]) {
+            setTimeout(destroy, 1300, arr[0], arr[1])
```
zonzujiro: `1300` - magic numbers are [bad](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)

Also, pls, replace passing of arguments into `setTimeout` with a function.

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,61 @@
...
+            return;
```
zonzujiro: Redundant `return` because of `if` statement.

zonzujiro:
Resolve outdated discussions

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,86 @@
...
+        gameCard.id = index;
+    }
+);
+
+document.querySelector('.main-wrapper').appendChild(wrapper);
+
+wrapper.addEventListener('click', function () {
+    let target = event.target.parentNode;
+        target = target.parentNode;
+    if(isNumeric(target.id)) main(target.id);
```
zonzujiro: And you need this check why?

submissions/dnzln/memory-game/app.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+);
+
+function main(id) {
+    click.push(id);
```
zonzujiro: `clickedCards` is better

submissions/dnzln/memory-game/app.js
dnzln: I thought, if click was on another object, not image, target.id will be incorrect and it's will crashing main function. I had numbers as ID for these divs, so I decided to make this check. But now it seems that checking on target.id thruthy could be enough.

submissions/dnzln/memory-game/app.js
dnzln: It is just array for checking in different if-statements, so I wished to keep it short to make code smaller and easy to read all these conditions. I got that, will name variables more clear.

submissions/dnzln/memory-game/app.js
zonzujiro: First of all, I am not sure if I told, but don't use constructions like `children[0]`, `parentNode`, `nextSibling`, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. 

Which is bad, obviously. Use `contains` instead. With `contains` you can find clicked item with `id`.

submissions/dnzln/memory-game/app.js
zonzujiro: I mean `click` - is action. But you have an array of some entities. It's just confusing right now.

zonzujiro:
Pls, resolve outdated stuff

dnzln:
I'm sorry, I just don't get what exactly I need to do right now. I see that PR was merged and closed but in the same time there are requested changes and these outdated stuff, which I don't know what they meen. Can you please explain?

zonzujiro:
All comments only FYI. In general, you made a good job so I have merged it.

PR #261, opened 2019-01-04
Friends app

KurosavaAkira:
### Friends App

[Demo](https://kurosavaakira.github.io/friends-app/) |
[Code base](https://github.com/KurosavaAkira/friends-app)

Please, review.


submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const sortVisibleToTop = () => {
+    friendsData.sort((a, b) => {
+        return b.visible - a.visible;
+    }); 
+}
+
+const inputEqualToFriendName = (e, friend) => {
+    if (friend.name.indexOf(e.target.value.toLowerCase()) >= 0) return true;
```
zonzujiro: `=> String#includes`

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,120 @@
...
+    return false;
+}
+
+const updateDom = () => {
```
zonzujiro: I see you have done a lot of work here, but right now you have a performance issue.

Pls, try to rewrite functionality which moves DOM elements using `translate` and `translate` - https://robots.thoughtbot.com/transitions-and-transforms#translate

You can read about performance bottlenecks in our optional task. Also, fill free to ask questions.

submissions/KurosavaAkira/friends-app/components/Friends.js
```diff
@@ -0,0 +1,60 @@
+import { friendsData } from './Data.js';
+
+const generateFriendsHtml = () => {
```
zonzujiro: Same with performance

submissions/KurosavaAkira/friends-app/components/Friends.js
```diff
@@ -0,0 +1,60 @@
...
+      blurNotSelectedFriends(friend.id, friend);
+    }
+}
+
+const toggleSelectedFriend = (friend) => {
+    const name_age_phone = document.querySelectorAll(`[id='${friend.id}'] > div:nth-last-of-type(-n+3)`);
+    friend.classList.toggle('friend-selected');
+    name_age_phone[0].classList.toggle('name-selected');
+    name_age_phone[1].classList.toggle('age-phone-visible');
+    name_age_phone[2].classList.toggle('age-phone-visible');
```
zonzujiro: I see this is a good case for array destructuring - `[name, age, phone]`

KurosavaAkira:
@zonzujiro Thanks a lot for comments! I made changes according to your notes.

BTW, transform and translate worked perfectly for performance optimization. It was a dramatic change in fps. I was surprised that browser can have such an issue with the transitions.

There was only one problem with your comment about `.sort` function:

> No need in `return` here. Just write `sort((a, b) => a - b)`

This function didn't work without the return statement.

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,120 @@
...
+    buttons.forEach( button => {
+        if (button.value === value) button.classList.add('sorting-buttons-active');
+        else button.classList.remove('sorting-buttons-active');
+    });
+}
+
+
+const addSearchEvent = () => {
+    const search = document.getElementById('search');
+    search.onkeyup = (e) => {
```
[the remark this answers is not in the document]
KurosavaAkira: > will keyup work if I paste value using a mouse?

The mouse event didn't work, so I added `input` eventListener.

submissions/KurosavaAkira/friends-app/components/Data.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+class Friend {
+    constructor(friend, id, visible = true) {
+        this.id = id;
+        this.photo = friend.picture.large;
+        this.name = `${friend.name.first} ${friend.name.last}`;
+        this.age = friend.dob.age;
+        this.phone = friend.cell;
+        this.visible = visible;
```
zonzujiro: Is this a boolean variable? If so, according to a naming convention, such stuff must be named as `isVisible`.

Names of boolean variables must start from `is`, `has`, `was`, etc.

submissions/KurosavaAkira/friends-app/components/Data.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+} 
+
+const createFriendsList = async () => {
+    const raw_friends_list = await getFriends();
```
zonzujiro: `camelCase` here also pls :)

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,153 @@
...
+    highlightActiveButton(e.target.value);
+    updateDom();
+    });
+}
+
+const sortAge = () => {
+    if (options.sort === 'age') {
+        options.sort = 'age-reverse';
+        friendsData.sort((a, b) => {
+            return b.age - a.age;
```
zonzujiro:
```javascript
friendsData.sort((a, b) => b.age - a.age)
```

😄

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,153 @@
...
+        options.sort = 'age';
+        friendsData.sort((a, b) => {
+            return a.age - b.age;
+        });
+    }
+}
+
+const sortReset = () => {
+    friendsData.sort((a, b) => {
+        options.sort = 'reset';
```
zonzujiro: I am not sure that this is a good idea.

First of all, sorting functions must be pure as a possible, without side-effects.
Second - you changing external variable during sorting. Obviously, this changes not part of sorting itself.
And last - you will do this changing during every iteration of the loop.

If you need to change this variable - move it outside  `Array#sort`.

Also, pls, can you create constants with sorting orders? Such stuff will make code more readable

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,153 @@
...
+            button.classList.add('sorting-buttons-active');
+            button.classList.remove('sorting-buttons-active-reverse')
+            break;
+        case 'age-reverse':
+        case 'name-reverse':
+            button.classList.add('sorting-buttons-active');
+            button.classList.add('sorting-buttons-active-reverse')
+            break;
+    }
+}
```
zonzujiro: `button.classList.add('sorting-buttons-active');` - you always do this operation. So maybe move it outside of `switch` statement? 🤔 

And then you can do like this:

```javascript
const sortingOptions = [SORT_BY_AGE, SORT_BY_NAME, RESET_SORTING]

button.classList.add('sorting-buttons-active');

if (sortingOptions.includes(options.sort) {
  button.classList.remove('sorting-buttons-active-reverse')
} else {
  button.classList.add('sorting-buttons-active-reverse')
}
```

Less lines of code, as you see.

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,153 @@
...
+
+const addSearchEvent = () => {
+    const search = document.getElementById('search');
+    let events = ['keyup', 'input'];
```
zonzujiro: You can use or `input` or `change`. No need in `keyup` and in using several events.

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,153 @@
...
+    let x = 0;
+    let y = -180;
+    for (let i = 0; i < friendsData.length; i++) {
+        if (i % items_in_row === 0) {
+        y += 180;
+        x = 0;
+        }
+        else x += 150;
+        const friend_node = document.getElementById(friendsData[i].id);
+        friend_node.style.transform = `translate(${x}px, ${y}px)`;
```
zonzujiro: 🔥

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,153 @@
...
+    if (friend.name.includes(e.target.value.toLowerCase())) return true;
+    return false;
+}
+
+const updateDom = () => {
+    const items_in_row = 7;
+    let x = 0;
+    let y = -180;
+    for (let i = 0; i < friendsData.length; i++) {
+        if (i % items_in_row === 0) {
```
zonzujiro: Such code parts usually hard to understand by rest developers. Can you provide some comments about logic of calculation, maybe? 🤔

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -0,0 +1,153 @@
...
+        y += 180;
+        x = 0;
```
zonzujiro: Indentation

submissions/KurosavaAkira/friends-app/components/Friends.js
```diff
@@ -0,0 +1,60 @@
...
+    const items_in_row = 7;
+    let x = 0;
+    let y = -180;
+    let html_string = '<div class="friends-list">';
+    for (let i = 0; i < FRIENDS_LIST.length; i++) {
+      if (i % items_in_row === 0) {
+        y += 180;
+        x = 0;
+      }
+      else x += 150;
```
zonzujiro: This part almost the same as in `updateDom`? Can it be combined into one function to keep code more DRY?

submissions/KurosavaAkira/friends-app/components/Friends.js
```diff
@@ -0,0 +1,60 @@
...
+                         <div class="age">Age: ${FRIENDS_LIST[i].age}</div>
+                         <div class="phone">Phone: ${FRIENDS_LIST[i].phone}</div>
+                      </div>`;
+    }
+    return html_string + '</div>';
+  }
+
+const addFriendEvent = () => {
+    const friends = document.getElementsByClassName('friends-list')[0];
+    friends.onclick = (e) => {
```
zonzujiro: Same with `addEventListener` - https://github.com/kottans/frontend-2019-homeworks/pull/261#discussion_r245357299

submissions/KurosavaAkira/friends-app/components/Friends.js
```diff
@@ -0,0 +1,60 @@
...
+                         <div class="name">${FRIENDS_LIST[i].name}</div>
+                         <div class="age">Age: ${FRIENDS_LIST[i].age}</div>
+                         <div class="phone">Phone: ${FRIENDS_LIST[i].phone}</div>
+                      </div>`;
+    }
+    return html_string + '</div>';
+  }
+
+const addFriendEvent = () => {
+    const friends = document.getElementsByClassName('friends-list')[0];
```
zonzujiro: Use `querySelector` here, because it will return one element.

submissions/KurosavaAkira/friends-app/components/Friends.js
```diff
@@ -0,0 +1,60 @@
...
+    friends.onclick = (e) => {
+      const friend = e.target.parentNode;
```
zonzujiro: Don't use constructions like `children[0]`, `parentNode`, `nextSibling`, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use `querySelector` or smth instead.

submissions/KurosavaAkira/friends-app/components/Filters.js
KurosavaAkira: Interesting solution, thanks. 

I thought switch statement was more readable and easier to understand for other programmers.

submissions/KurosavaAkira/friends-app/components/Friends.js
KurosavaAkira: I'm really not sure how to make this part of code universal and simple to easy to understand at the same time.

submissions/KurosavaAkira/friends-app/index.css
```diff
@@ -0,0 +1,223 @@
...
+    margin: 20px;
+    z-index: 1;
+    cursor: pointer;
+    transition: all 1.5s ease-in-out, filter 1s ease-in-out;
+}
+
+.friend-selected {
+    z-index: 200 !important;
+    transform: scale(3) !important;
+    transform: scale(3) translate(150px, 50px) !important;
```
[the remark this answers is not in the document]
lempiy: @KurosavaAkira You can use absolute CSS units for container and relative units for its content (like images and text). For instance https://css-tricks.com/fitting-text-to-a-container/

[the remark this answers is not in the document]
lempiy: up^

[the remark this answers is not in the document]
lempiy: https://css-tricks.com/scaling-responsive-animations/

submissions/KurosavaAkira/friends-app/index.css
```diff
@@ -0,0 +1,223 @@
...
+}
+
+.friend-not-selected {
+    pointer-events: none !important;
+    filter: blur(10px) grayscale(0.6) !important;
```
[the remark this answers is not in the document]
KurosavaAkira: @lempiy Is it ok to clone the clicked node in absolute, place it over the parent node and blur only one container of friends? Thus the clicked friend will disappear immediately and cloned will be zoomed to center.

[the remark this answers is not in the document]
lempiy: yes it may work. Also you don't need to clone node each time you click. You can create one `details` element and replace its content each time user picks new friend.

submissions/KurosavaAkira/friends-app/components/Filters.js
zonzujiro: Well, you have only one check - `if 'one of' OR 'else'`, so `Array#includes` appropriate and shorter. What if you will have more then 3 checks - imagine the amount of `switch` lines.

Also, at least from my experience, `switch` is a replacement for `else if` statement and extremely rare. Because if you have a lot if `else if` - something is wrong and the code, not abstract, which is bad. Such code harder to reuse because it has very concrete logic.

submissions/KurosavaAkira/friends-app/components/Friends.js
zonzujiro:
```
const recalculateCardPositions = friends => {
  const positions = [];
  const itemsInRow = 7;
  let x = 0;
  let y = -180;

  for (let i = 0; i < friends.length; i++) {
    // if there is place in a row - add item to row
    if (i % itemsInRow) {
      x += 150;

      // if no place in a row - move to new row
    } else {
      y += 180;
      x = 0;
    }

    positions.push({ x, y });
  }

  return positions;
};



const positions = recalculateCardPositions(friendsData);

positions.forEach(({ x, y }, i) => {
  const node = document.getElementById(friendsData[i].id);
  node.style.transform = `translate(${x}px, ${y}px)`;
});



const createFriendCard = (friend, { x, y }) => {
  const { id, photo, name, age, phone } = friend;

  return `<div class="friend" id="${id}" style="transform: translate(${x}px, ${y}px);">
            <div class="photo" style="background-image: url('${photo}')"></div>
            <div class="name">${name}</div>
            <div class="age">Age: ${age}</div>
            <div class="phone">Phone: ${phone}</div>
        </div>`;
};

const cards = positions
  .map((pos, i) => createFriendCard(friendsData[i], pos))
  .join('');
```

Something like this should work

submissions/KurosavaAkira/friends-app/components/Friends.js
zonzujiro: Yes, you can use `contains` - it will look for a child node with necessary attributes.

https://developer.mozilla.org/en-US/docs/Web/API/Node/contains.

Also, `onclick` - is a bad practice. 😉

KurosavaAkira:
@lempiy @zonzujiro Thanks a lot for comments. 
I made essential changes to code in the last commit. Please, review.

submissions/KurosavaAkira/friends-app/components/Data.js
```diff
@@ -16,21 +16,54 @@ const getFriends = async () => {
...
+// Each new items is generated from left to right. 
+// After every 7 items, there will be a new line (+180px).
+//
+//  y
+//  ↑
+//  ╎ (0:0)
+//  ╎  ▯▯▯▯▯▯▯ 
+//  ╎  ▯▯▯▯▯▯▯ 
+//  ╎  ▯▯▯▯▯▯▯
+//  -----------→x
```
zonzujiro: 😍 Great comment. I am serious.

submissions/KurosavaAkira/friends-app/components/Data.js
```diff
@@ -16,21 +16,54 @@ const getFriends = async () => {
...
+    const positions = [];
+    const itemsInRow = 7;
+    let x = 0;
+    let y = -180;
```
zonzujiro: Like an idea, you can save 180 into variable `LINE_HEIGHT` and 150 into `CARD_WIDTH`

submissions/KurosavaAkira/friends-app/index.css
lempiy: Nice, I've got FPS increasement  on my 'toaster'. Now its about 10fps average.
Use Performance Tab on Dev Tools to debug performance issues. See bellow.
The main problem with your animation is flowing:
1. Frequent layer changes. (transition on z-index isn't so good you know :))
2. Verbose style recalculation.
3. You're still animating tiles one-by-one.
screen shot 2019-01-10 at 9 26 15 am

Actually your animation is easy to implement but way not so easy to make "production-ready". You may leave it as it is, since reaching the goal maybe a long process. But if you have time I have suggestion to try. 

My suggestion:
Move selected and non-selected tiles to separate overlaying containers (`friends-list` (z-index: 1), `friends-details` (z-index: 2)) that will act as `Photoshop Layers`.
Your `friends-details` layer will be on top with only one element hidden - `.seleted-friend`.
Upon clicking on non-selected friend following happens:
1. `.seleted-friend` content (text, image) will be set to non-selected friends content.
2. `non-selected` will get `visability: hidden`
3. `seleted-friend` will become visible and "replace" hidden `non-selected` by applying its `transform(x, y)` position.
4. `non-selected` ARE NOT going to be animated and they will not have transitions. We are going to animate their container `friends-list` (all layer).
5. Animate `.seleted-friend` and `friends-list` with transitions (scale, filter etc).

With such approach we may reduce z-index mess and will leave browser more space for optimization.
You can also try to add `will-change` property to `friends-list` container or `.seleted-friend`. It may increase performance, but will depend on your composite layers this should be tested using DEV tools.

submissions/KurosavaAkira/friends-app/components/Filters.js
```diff
@@ -81,34 +76,27 @@ const highlightActiveButton = (value) => {
...
+        button.classList.remove('sorting-buttons-active-reverse');
+    } else {
+        button.classList.add('sorting-buttons-active-reverse');
     }
 }
 
 const addSearchEvent = () => {
     const search = document.getElementById('search');
-    let events = ['keyup', 'input'];
+    let events = ['input'];
```
zonzujiro: I believe you don't need an array for one entity.

So `search.addEventListener('input', ...` should be ok.

submissions/KurosavaAkira/friends-app/components/Friends.js
```diff
@@ -1,34 +1,33 @@
...
-      toggleSelectedFriend(friend);
-      blurNotSelectedFriends(friend.id, friend);
-    }
+    const friends = document.querySelector('.friends-list');
+    friends.addEventListener('click', (e) => {
+      const friend = e.target.closest('.friend');
+      if (friend === null) return;
+      else {
+        toggleSelectedFriend(friend);
+        blurNotSelectedFriends(friend.id, friend);
```
zonzujiro: No need in `friend.id` if you pass all `friend` already.

submissions/KurosavaAkira/friends-app/components/Friends.js
```diff
@@ -39,21 +38,25 @@ const toggleSelectedFriend = (friend) => {
     phone.classList.toggle('age-phone-visible');
 }
 
-const blurNotSelectedFriends = (selected_id) => {
+const blurNotSelectedFriends = (selectedId) => {
+    const dark_layer = document.querySelector('.dark-layer');
```
zonzujiro: `camelCase` pls

zonzujiro:
image

image

Okey... So I will merge it

submissions/KurosavaAkira/friends-app/index.css
lempiy: screen shot 2019-01-10 at 10 14 50 am

PR #268, opened 2019-01-04
Memory pair game

olek557:
### Memory – Pair Game

[Demo](https://olek557.github.io/memory-game/) |
[Code base](https://github.com/olek557/memory-game)

Please, review.


submissions/olek557/memory-pair-game/script.js
```diff
@@ -0,0 +1,57 @@
+// BOARD CLASS
+let Board = function(id) {
+  this.sequence = [];
+  this.boardElement = document.getElementById('board');
+}
+
+Board.prototype.generate = function() {
+  for(i = 0; i < 8; i++) {
```
AMashoshyna: This method is doing too many things: it generates initial card sequence, it shuffles the cards, adds event handler to the board and event contains the logic of comparing the cards. Please refactor it into small reusable functions which do only one thing each.

submissions/olek557/memory-pair-game/script.js
```diff
@@ -0,0 +1,57 @@
...
+    this.sequence.push(new Card(i), new Card(i));
+  }
+  this.sequence.sort(function() { return 0.5 - Math.random() });
+  this.boardElement.addEventListener('click', event => {
+    if(event.target.matches('.card__front')) {
+      let cardElement = event.target.parentNode,
+          index = cardElement.getAttribute('data-cell');
+      cardElement.parentNode.classList.add('open');
+      cardElement.querySelector('.card__back').innerHTML = this.sequence[index].value;
+      if(selectedCard) {
```
AMashoshyna: `selectedCard` is a global variable, which are source of many bugs. It is better to add it to board's `this` next to sequence.

AMashoshyna:
Either I'm very lucky, or card shuffling algorithm needs some improvement =)

image

image


From the point of view of user experience I would suggest to either hide matching cards or at least mark them as discarded somehow. Greyish color or transparency maybe? If you don't mark them in any way, they are very distracting.

To the code:
This piece of html is repeated 16 times:
image

Do you think you could put it in js to reduce code?
Also, please see my comments to code.

olek557:
@AMashoshyna Thanks a lot for review. I changed almost all things in PR. Only array shuffling i didn't change,( i got this code from article which provided in additions in article). I tested on my own and it looks better than on your's screens.

olek557:
Got it. Deleted this.

AMashoshyna:
Well done! LGTM.

PR #269, opened 2019-01-04
DOM task

vv2529:
### DOM
[Demo](https://vv2529.github.io/gh-pages-js-dom/)
[Code base](https://github.com/vv2529/kottans-frontend/tree/master/task_js_dom/js-dom)

Please, review.


submissions/vv2529/js-dom/index.html
```diff
@@ -0,0 +1,21 @@
...
+			<li class="menu-item">commit</li>
+			<li class="menu-item">hoist</li>
+			<li class="menu-item">traverse</li>
+			<li class="menu-item">mitigate</li>
+		</nav>
+		<main id="main"></main>
+		<button id="navbar-toggle">☰</button>
+		<script src="script.js"></script>
+	</body>
+</html>
```
AMashoshyna: This red arrow means that you need to add newlines at the end of every file:

image

submissions/vv2529/js-dom/script.js
```diff
@@ -0,0 +1,61 @@
+// "Id" is much easier to write :)
+function Id(id){return document.getElementById(id)}
```
AMashoshyna: I really like your approach! Creating useful helper functions is nice. But two points to be considered:
- naming: functions should have verbs in their names (exceptions: word like "action" or "handler", which imply action). Also only classes and constructor function names can start with uppercase letters.
- DOM queries are expensive. So if you need to query an element more than one time, better put in in a variable and refer to this variable.

submissions/vv2529/js-dom/script.js
```diff
@@ -0,0 +1,61 @@
...
+	loadData(title);
+	Id('navbar').classList.remove('shown');
+});
+
+/*
+ * Functionality
+ */
+
+let currentPage = '';
+const savedData = {};
```
AMashoshyna: Nice idea with lazy loading and caching!

submissions/vv2529/js-dom/script.js
```diff
@@ -0,0 +1,61 @@
...
+		currentPage = title;
+	}
+	else showError();
+	console.log('displayData:', title);
+}
+
+function loadData(title){
+	// Check if requested data is cached
+	if(title in savedData) displayData(title);
+	else fetch('data/'+title+'.json')
```
AMashoshyna: Try template literals here.

submissions/vv2529/js-dom/script.js
```diff
@@ -0,0 +1,61 @@
...
+
+function showError(title){
+	Id('main').innerHTML = '<h1>No content!</h1><p>No data could be displayed for <q>'+title+'</q> query, sorry :(</p>';
+}
+
+/*
+ * Initiate
+ */
+
+loadData('home');
```
AMashoshyna: A function should do only one thing. `loadData` loads the data AND displays it. Please refactor.

submissions/vv2529/js-dom/data/commit.json
```diff
@@ -0,0 +1,3 @@
+{
+	"content": "<h1>Commit</h1><p>commit /kəˈmɪt/ (<i>verb</i>) —<ol><li><i>(CRIME)</i> to do something illegal or something that is considered wrong.<ul class=\"example-list\"><li>He was sent to prison for a crime that he didn't commit.</li><li>to commit adultery/murder</li><li>to commit an offence</li></ul></li><li><i>(PROMISE)</i> to promise or give your loyalty, time, or money to a particular principle, person, or plan of action.<ul class=\"example-list\"><li>Like so many men, he has problems committing himself to a relationship.</li><li>The government must commit itself to improving healthcare.</li><li>Once we have committed to this course of action there is no going back.</li></ul></li><li><i>(SEND)</i> to send someone officially to prison or hospital.<ul class=\"example-list\"><li>He's been committed to prison for fraud.</li></ul></li></ol></p>"
```
AMashoshyna: Please don't mix data and representation. Normally you would receive the data from server. If your layout changes due to redesign, you would need to introduce changes into both frontend and backend code, which is bad. In short: try to separate data(texts in this case) from representation details (html tags here) as much as possible.

AMashoshyna:
Very good job!

PR #270, opened 2019-01-04
js-dom task done

mfialko:
### js-dom

[Demo](https://mfialko.github.io/jsdom-task/) |
[Code base](https://github.com/mfialko/jsdom-task)

Please, review.


submissions/mfialko/js-dom/app.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+var content = [
+    { 'text': "<h2>Kottans Front-End</h2><p>Click on the menu button to change the content inside the main section.</p>",
```
AMashoshyna: Please don't mix data and representation. Normally you would receive the data from server. If your layout changes due to redesign, you would need to introduce changes into both frontend and backend code, which is bad. In short: try to separate data(texts in this case) from representation details (html tags here)  as much as possible.

submissions/mfialko/js-dom/app.js
```diff
@@ -0,0 +1,37 @@
...
+    { 'text': "<h2>Verear noluisse intellegebat eu nec</h2><p>Eu sit utinam ubique, ea ius ubique mediocrem. Vis id inermis verterem, sea et viderer consetetur definitiones. Quo cibo interpretaris ei, platonem dissentias vim ex, ancillae iracundia eos et.</p><p>Lorem maiormum ut per, dicta nostro labores mel ex, eos molestie definitiones cu. Sit an augue dolore nonumy, ad nostrud intellegam per.Nam et adolescens incorrupte, ex nulla mazim utroque nec. At labores appetere nam. Mel ea option dolorum, usu clita epicurei volutpat eu, iuvaret deseruisse ex eam. Ut cibo evertitur rationibus eam, quodsi ancillae maluisset mel eu.</p>"}
+]; 
+
+const sideMenu = document.querySelector('.side-menu');
+const main = document.querySelector('main'); 
+const mobileMenu = document.querySelector('#menu')
+
+for (let i=0; i<5; i++) {
+
+    sideMenu.children[i].addEventListener('click', function() {
```
AMashoshyna: Please use event delegation instead of adding event listeners to every single item in collection. Add one listener to container element instead.

submissions/mfialko/js-dom/index.html
```diff
@@ -0,0 +1,49 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Practical task</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" type="text/css" href="style.css">
+  </head>
+  <body>
+
```
AMashoshyna: Please remove redundant empty lines (here and everywhere applicable).

submissions/mfialko/js-dom/index.html
```diff
@@ -0,0 +1,49 @@
...
+          <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+            <path d="M2 6h20v3H2zm0 5h20v3H2zm0 5h20v3H2z"/>
+          </svg>
+        </a>
+      </div>
+    </header>
+        
+      
+      <nav class="side-menu">
+         <a href="#about">About</a>
```
AMashoshyna: Why did you decide to make menu with anchor elements? Just curious about your reasoning.

mfialko: I thought that easier to manipulate anchor elements like buttons for menu items, than list elements. I saw tutorials about and it seemed as a usual practice.

submissions/mfialko/js-dom/app.js
mfialko: Thank you. At first I wanted to separate tags from text, but then I hesistate about different numbers of paragraps in sections and, in purpose of task, I decided to leave it.

submissions/mfialko/js-dom/index.html
AMashoshyna: One of unwanted side-effects is url change, which causes browser's "back" button to act weird: it looks like you can go back to some previous state, but nothing happens. In short, it is better to choose other html element and provide necessary style with css (use :hover to control cursor appearance etc.)

submissions/mfialko/js-dom/app.js
mfialko: Updated.

submissions/mfialko/js-dom/index.html
mfialko: Done.

submissions/mfialko/js-dom/app.js
mfialko: Updated

AMashoshyna:
Well done! I will merge the PR, but please don't use `a` tag when there is no actual link. There are more appropriate elements for list items or navigation. Also `a` element without `href` attribute is invalid HTML.

PR #273, opened 2019-01-04
create friends app

yulyasystem:
### friends-api-task

[Demo](https://yulyasystem.github.io/friend-app/) |
[Code base](https://github.com/yulyasystem/friend-app)

Please, review.

### DIFF
--- a/submissions/yulyasystem/friends-app/script.js
+++ b/submissions/yulyasystem/friends-app/script.js
@@ -5,7 +5,8 @@ let sortSpan = document.querySelector('.sort');
 let filter = document.querySelector('.filter');
 let grid = document.querySelector('.grid');
 let search = document.querySelector('.search');
-let isSorted = false;
+let arrowDown = document.querySelector('.arrow-down');
+let arrowUp = document.querySelector('.arrow-up');
 
 function handleErrors(response) {
     if (!response.ok) {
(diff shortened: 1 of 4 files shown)


submissions/yulyasystem/friends-app/script.js
```diff
@@ -0,0 +1,112 @@
...
+            </div>`;
+    });
+    container.appendChild(div);
+}
+
+function sortByAge(data) {
+
+
+    let grid = document.querySelector('.grid');
+    let sortedByAge = data.sort((a, b) => {
```
zonzujiro: You don't need to use `return`, instead, you can do like that - `data.sort((a,b) => a - b)`.
Also, `data` is to general name. Can, replace it with something concrete?

submissions/yulyasystem/friends-app/script.js
```diff
@@ -0,0 +1,112 @@
...
+        return a.dob.age - b.dob.age;
+    }).slice();
+    let sortedByAgeDown = data.sort((a, b) => {
+        return b.dob.age - a.dob.age;
+    }).slice();
```
zonzujiro: As far as I can see it, you always sort your array in two orders, which is redundant. 
Can you sort an array according to order which was chosen by a user? I mean without `isSorted`.

submissions/yulyasystem/friends-app/script.js
```diff
@@ -0,0 +1,112 @@
...
+    let grid = document.querySelector('.grid');
+    let searchArray = [];
+    data.forEach(item => {
+        if (item.name.first === input) {
+            console.log(item);
+            searchArray.push(item);
+        }
+    });
+    grid.innerHTML = "";
+    if (searchArray.length !== 0) {
```
zonzujiro: You can simplify this check to `seachArray.length`

submissions/yulyasystem/friends-app/script.js
```diff
@@ -0,0 +1,112 @@
...
+
+}
+
+function searchByName(input, data) {
+    let grid = document.querySelector('.grid');
+    let searchArray = [];
+    data.forEach(item => {
+        if (item.name.first === input) {
+            console.log(item);
+            searchArray.push(item);
```
zonzujiro: You can use `Array#map` instead.

```
const seachArray = data.map(....
```

PR #274, opened 2019-01-05
oop exercise

sioniks:
### OOP exercise

[Demo](https://sioniks.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/sioniks/kottans_frontend/tree/master/task_9/)

Please, review.

### DIFF
--- a/submissions/sioniks/oop-exercise/index.js
+++ b/submissions/sioniks/oop-exercise/index.js
@@ -8,57 +8,87 @@
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
 class Inhabitant {
-  constructor(name, gender, saying, friends) { //, species, legs, hands
+  constructor(name, gender, saying, friends) {
     this.name = name;
     this.gender = gender;
     this.saying = saying;
     this.friends = friends;
-    // this.species = species;
-    // this.legs = legs;
-    // this.hands = hands;
   }
 
+  getInfo() {}
+}
+
+class Cat extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super(name, gender, saying, friends);
+    this.species = 'cat';
+    this.legs = 4;
+  }
   getInfo() {
+    super.getInfo();
     let massage = [
       `${this.species}`,
       `<strong>${this.name}</strong>`,
       `${this.gender}`,
-      this.legs === (null || undefined) ? '' : `${this.legs}`,
-      this.hands === (null || undefined) ? '' : `${this.hands}`,
+      `${this.legs}`,
       `<em>${this.saying}</em>`,
       `${this.friends}`
     ];
     massage = massage.filter(function (n) {
-      return n != ''
+      return n !== ''
     });
     return massage.join('; ');
   }
 }
 
-class Cat extends Inhabitant {
-  constructor(name, gender, saying, friends, species = 'cat', legs = 4) {
-    super(name, gender, saying, friends);
-    this.species = species;
-    this.legs = legs;
-  }
-}
-
 class Dog extends Inhabitant {
-  constructor(name, gender, saying, friends, species = 'dog', legs = 4) {
+  constructor(name, gender, saying, friends) {
     super(name, gender, saying, friends);
-    this.species = species;
-    this.legs = legs;
+    this.species = 'dog';
+    this.legs = 4;
+  }
+  getInfo() {
+    super.getInfo();
+    let massage = [
+      `${this.species}`,
+      `<strong>${this.name}</strong>`,
+      `${this.gender}`,
+      `${this.legs}`,
+      `<em>${this.saying}</em>`,
+      `${this.friends}`
+    ];
+    massage = massage.filter(function (n) {
+      return n !== ''
+    });
+    return massage.join('; ');
   }
 }
 
 class Human extends Inhabitant {
-  constructor(name, gender, saying, friends, species = 'human', legs = 2, hands = 2) {
+  constructor(name, gender, saying, friends) {
     super(name, gender, saying, friends);
-    this.species = species;
-    this.legs = legs;
-    this.hands = hands;
+    this.species = 'human';
+    this.legs = 2;
+    this.hands = 2;
+  }
+  getInfo() {
+    super.getInfo();
+    let massage = [
+      `${this.species}`,
+      `<strong>${this.name}</strong>`,
+      `${this.gender}`,
+      `${this.legs}`,
+      `${this.hands}`,
+      `<em>${this.saying}</em>`,
+      `${this.friends}`
+    ];
+    massage = massage.filter(function (n) {
+      return n !== ''
+    });
+    return massage.join('; ');
   }
 }
+
 const dog = new Dog('Toby', 'male', 'woof-woof!', ['Tiki, Taki']);
 const cat = new Cat('Fabula', 'female', 'meow-meow!', ['Tom, Mimi']);
 const man = new Human('Kan', 'male', 'I am Superman!', ['Tony, Alice']);
@@ -90,4 +120,3 @@ inhabitants.forEach(element => {
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
    */
-  
--- a/submissions/sioniks/oop-exercise/index.js
+++ b/submissions/sioniks/oop-exercise/index.js
@@ -8,37 +8,59 @@
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
 class Inhabitant {
-  constructor(name, gender, saying, friends, species, legs, hands) {
-    this.species = species;
+  constructor(name, gender, saying, friends) { //, species, legs, hands
     this.name = name;
     this.gender = gender;
     this.saying = saying;
     this.friends = friends;
-    this.legs = legs;
-    this.hands = hands;
+    // this.species = species;
+    // this.legs = legs;
+    // this.hands = hands;
   }
 
   getInfo() {
-    let legsCount = this.legs === null ? '' : `${this.legs}; `;
-    let handsCount = this.hands === null ? '' : `${this.hands}; `;
-    return `${this.species}; ` + `<strong>${this.name}</strong>; ` + `${this.gender}; ` + legsCount + handsCount + `<em>${this.saying}</em>; ` + `${this.friends};`;
+    let massage = [
+      `${this.species}`,
+      `<strong>${this.name}</strong>`,
+      `${this.gender}`,
+      this.legs === (null || undefined) ? '' : `${this.legs}`,
+      this.hands === (null || undefined) ? '' : `${this.hands}`,
+      `<em>${this.saying}</em>`,
+      `${this.friends}`
+    ];
+    massage = massage.filter(function (n) {
+      return n != ''
+    });
+    return massage.join('; ');
   }
 }
 
-class Animal extends Inhabitant {
-  constructor(name, gender, saying, friends, species, legs = 4, hands = null) {
-    super(name, gender, saying, friends, species, legs, hands)
+class Cat extends Inhabitant {
+  constructor(name, gender, saying, friends, species = 'cat', legs = 4) {
+    super(name, gender, saying, friends);
+    this.species = species;
+    this.legs = legs;
   }
+}
 
+class Dog extends Inhabitant {
+  constructor(name, gender, saying, friends, species = 'dog', legs = 4) {
+    super(name, gender, saying, friends);
+    this.species = species;
+    this.legs = legs;
+  }
 }
 
 class Human extends Inhabitant {
-  constructor(name, gender, saying, friends, species = 'human', legs = null, hands = 2) {
-    super(name, gender, saying, friends, species, legs, hands)
+  constructor(name, gender, saying, friends, species = 'human', legs = 2, hands = 2) {
+    super(name, gender, saying, friends);
+    this.species = species;
+    this.legs = legs;
+    this.hands = hands;
   }
 }
-const dog = new Animal('Toby', 'male', 'woof-woof!', ['Tiki, Taki'], 'dog');
-const cat = new Animal('Fabula', 'female', 'meow-meow!', ['Tom, Mimi'], 'cat');
+const dog = new Dog('Toby', 'male', 'woof-woof!', ['Tiki, Taki']);
+const cat = new Cat('Fabula', 'female', 'meow-meow!', ['Tom, Mimi']);
 const man = new Human('Kan', 'male', 'I am Superman!', ['Tony, Alice']);
 const woman = new Human('Kandy', 'female', 'Who is there?', ['Kan, Andy']);
 const catWoman = new Human('Lili', 'female', 'meow-meow! Who is there?', ['Tor, Iron-man']);
@@ -48,6 +70,7 @@ let inhabitants = [dog, cat, woman, man, catWoman];
 inhabitants.forEach(element => {
   print(element.getInfo());
 });
+
 
 // ======== OUTPUT ========
 /* Use print(message) for output.
@@ -67,3 +90,4 @@ inhabitants.forEach(element => {
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
    */
+  
--- a/submissions/sioniks/oop-exercise/index.js
+++ b/submissions/sioniks/oop-exercise/index.js
@@ -8,24 +8,16 @@
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
 class Inhabitant {
-  constructor(name, gender, saying, friends) {
+  constructor(name, gender, saying, friends, species, legs) {
     this.name = name;
     this.gender = gender;
     this.saying = saying;
     this.friends = friends;
+    this.legs = legs;
+    this.species = species;
   }
 
-  getInfo() {}
-}
-
-class Cat extends Inhabitant {
-  constructor(name, gender, saying, friends) {
-    super(name, gender, saying, friends);
-    this.species = 'cat';
-    this.legs = 4;
-  }
   getInfo() {
-    super.getInfo();
     let massage = [
       `${this.species}`,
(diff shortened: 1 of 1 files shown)


submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,69 @@
...
+    this.friends = friends;
+    this.legs = legs;
+    this.hands = hands;
+  }
+
+  getInfo() {
+    let legsCount = this.legs === null ? '' : `${this.legs}; `;
+    let handsCount = this.hands === null ? '' : `${this.hands}; `;
+    return `${this.species}; ` + `<strong>${this.name}</strong>; ` + `${this.gender}; ` + legsCount + handsCount + `<em>${this.saying}</em>; ` + `${this.friends};`;
+  }
```
OleksiyRudenko: A hint: 
```
[
  varA, 
  "text", 
  `some text: ${varB}`,
  condition ? valueIfTrue : valueIfFalse,
].join("delimeter ")
```

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,69 @@
...
+class Inhabitant {
+  constructor(name, gender, saying, friends, species, legs, hands) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.saying = saying;
+    this.friends = friends;
+    this.legs = legs;
+    this.hands = hands;
+  }
```
OleksiyRudenko: Classes for species that do not have hands by natural design should not consequently have `hands` or any equivalent property.

The above also means that `getInfo` should not add anything about hands to the output string.

You may wonder how to manage info re hands. Check the comment to `Human` below.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,69 @@
...
+    super(name, gender, saying, friends, species, legs, hands)
+  }
+
+}
+
+class Human extends Inhabitant {
+  constructor(name, gender, saying, friends, species = 'human', legs = null, hands = 2) {
+    super(name, gender, saying, friends, species, legs, hands)
+  }
+}
```
OleksiyRudenko: Define `getInfo()` it would call parent's method (`super.getInfo()`) and supplement it with info about hands.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+class Animal extends Inhabitant {
+  constructor(name, gender, saying, friends, species, legs = 4, hands = null) {
+    super(name, gender, saying, friends, species, legs, hands)
+  }
+
+}
+
+class Human extends Inhabitant {
+  constructor(name, gender, saying, friends, species = 'human', legs = null, hands = 2) {
```
OleksiyRudenko: By design and by default humans have 2 legs. `null` confuses.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,69 @@
...
+    super(name, gender, saying, friends, species, legs, hands)
+  }
+}
+const dog = new Animal('Toby', 'male', 'woof-woof!', ['Tiki, Taki'], 'dog');
+const cat = new Animal('Fabula', 'female', 'meow-meow!', ['Tom, Mimi'], 'cat');
+const man = new Human('Kan', 'male', 'I am Superman!', ['Tony, Alice']);
+const woman = new Human('Kandy', 'female', 'Who is there?', ['Kan, Andy']);
+const catWoman = new Human('Lili', 'female', 'meow-meow! Who is there?', ['Tor, Iron-man']);
```
OleksiyRudenko: `Human` is a species, `Animal` is a set of species. What's wrong with dogs and cats that they do not deserve their own classes? Please, eliminate discrimination.

submissions/sioniks/oop-exercise/index.js
OleksiyRudenko: Do we need to ask a user for species when we define a class that denotes a species?
You can call `super` with some explicit literal parameters just like any other function.

submissions/sioniks/oop-exercise/index.js
sioniks: done

submissions/sioniks/oop-exercise/index.js
sioniks: i fix it

sioniks:
@OleksiyRudenko please review now

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,93 @@
...
+/* Print examples:
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+
```
OleksiyRudenko: Your git is probably configured to keep OS specific newline code (Windows in this case).
This causes compatibility issues and unnecessary changes when collaborating with others.
It is important that newlines get converted to Linux style on commit and may convert back to OS specific on checkout.
Later, please, find a way to change git settings so it follows the workflow above re newlines.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,93 @@
...
+    // this.hands = hands;
+  }
+
+  getInfo() {
+    let massage = [
+      `${this.species}`,
+      `<strong>${this.name}</strong>`,
+      `${this.gender}`,
+      this.legs === (null || undefined) ? '' : `${this.legs}`,
+      this.hands === (null || undefined) ? '' : `${this.hands}`,
```
OleksiyRudenko: As per OOP principles a class cannot work with props that belong to child classes. That's why we need to use methods overloading (in `Human`).

OleksiyRudenko:
@sioniks Do you need any support on the above?

sioniks:
@OleksiyRudenko so I need support, but now I have no time for it. And I will be able to return to the task only tonight. Thank you for your support. I will describe my questions later in the comments.

submissions/sioniks/oop-exercise/index.js
sioniks: I understand that I have to do it this way
`getInfo() {
    super.getInfo();
    let massage = [
      `${this.species}`,
      `<strong>${this.name}</strong>`,
      `${this.gender}`,
      `${this.legs}`,
      `${this.hands}`,
      `<em>${this.saying}</em>`,
      `${this.friends}`
    ];
    massage = massage.filter(function (n) {
      return n !== ''
    });
    return massage.join('; ');
  }`

sioniks: @OleksiyRudenko I understand that I have to do the method in a superclass. and then supplement it with information on the hands and feet in other classes. But will not I repeat a lot of code?

sioniks: can I just take a class out function `getInfo()` and just call it later

sioniks:
I am not sure that I understood all the comments correctly. Please check this code for solving this problem.

OleksiyRudenko:
> I am not sure that I understood all the comments correctly. Please check this code for solving this problem.

When unsure, please, ask questions specifying (e.g. quoting) parts of comments that aren't clear. Otherwise it is very difficult to render an appropriate help.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+  constructor(name, gender, saying, friends) {
+    this.name = name;
+    this.gender = gender;
+    this.saying = saying;
+    this.friends = friends;
+  }
```
OleksiyRudenko: Some species do not have hands, but all species have legs and species properties.
There is no good reason to remove those from the base class.

> Aren't species and legs something all inhabitants have?
> Shared properties must be defined in a base class for to class hierarchy be OOP compliant.

Please, give some reasoning on your decision to remove those. This would help me to render somewhat better help.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+  getInfo() {}
+}
+
+class Cat extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super(name, gender, saying, friends);
```
OleksiyRudenko: `Cat.constructor` and `Inhabitant.constructor` (`super` gets resolved to `Inhabitant.constructor` here) not necessarily gotta have exactly same signatures ([function signature](https://developer.mozilla.org/en-US/docs/Glossary/Signature/Function)).

Why do you want to keep them similar?

Any function can accept literals (e.g. `"some string"` is a string literal, `55` is a numeric literal) along with variables.

Please, respond to a question above, for it would help to render somewhat better help.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,122 @@
...
+// Define your objects here
+class Inhabitant {
+  constructor(name, gender, saying, friends) {
+    this.name = name;
+    this.gender = gender;
+    this.saying = saying;
+    this.friends = friends;
+  }
+
+  getInfo() {}
```
OleksiyRudenko: This method is expected to return a string built from properties that `Inhabitant` owns.
Please, implement.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,122 @@
...
+}
+
+class Cat extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super(name, gender, saying, friends);
+    this.species = 'cat';
+    this.legs = 4;
+  }
+  getInfo() {
+    super.getInfo();
```
OleksiyRudenko: This call would make sense if it returned anything.

> Define `getInfo()` it would call parent's method (`super.getInfo()`) and supplement it with info about hands.

> @OleksiyRudenko I understand that I have to do the method in a superclass. and then supplement it with information on the hands and feet in other classes. But will not I repeat a lot of code?

You will not repeat a lot of code. Vice versa.

1. Have the base class returning a string built from props it owns.
1. In a child class method:
 - call base class method and take its result
 - add something (props) specific to the child class
 - return the aggregated result

E.g.
```
getInfo() {
  let d = super.getInfo(); //  - call base class method and take its result
  d = d + this.someProperty; // add something (props) specific to the child class
  return d; // return the aggregated result
}
```

Same fragment with comments based on an earlier comment
> Define `getInfo()` it would call parent's method (`super.getInfo()`) and supplement it with info about hands.
```
getInfo() { // Define getInfo()
  return  // (not mentioned in a comment above but otherwise this method makes no sense)
    super.getInfo() //  it would call parent's method (super.getInfo()) 
   + this.someProperty; // and supplement it with info about hands.
}
```

Please, ask a question if anything above is not clear.

OleksiyRudenko:
@sioniks 
> Please, also resolve conversations you believe you have elaborated on (unless there is a question). The above is an important part of communication, especially while learning.

submissions/sioniks/oop-exercise/index.js
OleksiyRudenko: @sioniks 
> Please, respond to a question above, for it would help to render somewhat better help.

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,93 @@
...
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+  constructor(name, gender, saying, friends) { //, species, legs, hands
+    this.name = name;
+    this.gender = gender;
+    this.saying = saying;
+    this.friends = friends;
+    // this.species = species;
+    // this.legs = legs;
```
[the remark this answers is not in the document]
sioniks: i fix it

submissions/sioniks/oop-exercise/index.js
sioniks: thanks for explanation

submissions/sioniks/oop-exercise/index.js
sioniks: I understood my mistake and corrected it

submissions/sioniks/oop-exercise/index.js
OleksiyRudenko: @sioniks Well, do not take print into consideration at this point (in general, it might never happen).

`Cat.constructor` expects `(name, gender, saying, friends)`.
`Cat.constructor` calls `super` with `(name, gender, saying, friends)`.
But super is actually a call to `Inhabitant.constructor` which expects `(name, gender, saying, friends, species, legs)`.
You just need to supplement `super` with remaining required arguments. These may be literals (`"text"` is a string literal, `55` is a numeric literal).

Does the above clarify things?

sioniks: this means that I have to transfer all the properties of the super in this case ?
`class Cat extends Inhabitant {`
 ` constructor(name, gender, saying, friends, species, legs) {`
  `  super(name, gender, saying, friends, species, legs);`
    `this.species = 'cat';`
    `this.legs = 4;`

and in the class man only add `hands` props like this

` constructor(name, gender, saying, friends, species, legs, hands) {`
  `  super(name, gender, saying, friends, species, legs, hands);`
`   this.hands = 2;`

OleksiyRudenko: @sioniks Not exactly, you have to **supply** the super with all required props. Not necessarily you need to take those from the user.
`super(varA, varB, "string", 5678)` is legal.
Just a question to think over: if you call a function from inside another function should their parameters be the same by all means?

sioniks: so props should be the same, otherwise i will get an error `props is not defined` if i make like this
`constructor(name, gender, saying, friends) {`
  `  super(name, gender, saying, friends, species, legs);`

if I make code 
`constructor(name, gender, saying, friends, species, legs) {`
  `  super(name, gender, saying, friends);`
`this.species = 'cat';`
`this.legs = 4;`

super method return whit default props

but I need call super constructor with all props that they have? right?

OleksiyRudenko: > so props should be the same, otherwise i will get an error `props is not defined` if i make like this
> `constructor(name, gender, saying, friends) {`
> ` super(name, gender, saying, friends, species, legs);`

Of course, since you do not have variables `species` and `legs`.

Doesn't the hint below help you?
> super(varA, varB, "string", 5678) is legal.

Try to think a bit out of the box.

My next step would be to give you an exact fragment of code, which I would like to avoid, as this will not have a learning value.

> but I need call super constructor with all props that they have? right?
You have to call `super` with all props the **super class requires**. It's a bit different from what you suggest.

OleksiyRudenko: Given the following code:
```
function doWhatever(name) {
  // whatever code
}

function doSomethingElse(familyName) {
  // whatever code
}
```

what would you do if I ask you to call `doSomethingElse` inside `doWhatever` with a family name, say, "Smith"?

NB! The above doesn't relate to the PR, OOP or anything within the code in the PR per se.

sioniks: in this code i can call `function doSomethingElse(name)` inside `function doWhatever`

OleksiyRudenko: > in this code i can call `function doSomethingElse(name)` inside `function doWhatever`

Where would you take name provided that we cannot change `doWhatever` function signature (i.e. arguments) and we also cannot take this from globals?

sioniks: ok in this example, I would just call a function `doSomethingElse` with props `"Smith"`

OleksiyRudenko: Right.
Now back to https://github.com/kottans/frontend-2019-homeworks/pull/274#discussion_r247086356

Any ideas or questions?

sioniks: realy??? I can do something like this

> constructor(name, gender, saying, friends) {
super(name, gender, saying, friends, 'dog', 4);

sioniks:
@OleksiyRudenko please review, I hope that at this moment, I will understand your tips right

submissions/sioniks/oop-exercise/index.js
OleksiyRudenko: What would prevent you from doing so?
This is a semi-rhetoric question. Reflect on that. You may optionally share your reflections here.

sioniks: I just thought that in constructor I can send only defined properties. And completely missed the moment that this is a function

OleksiyRudenko: Let me know, when you fix the code being armed with this learning.
If you have any further questions, please, do not hesitate addressing those to me.

sioniks: in last commit i fix code with this learning

submissions/sioniks/oop-exercise/index.js
```diff
@@ -0,0 +1,90 @@
...
+const cat = new Cat('Fabula', 'female', 'meow-meow!', ['Tom, Mimi']);
+const man = new Human('Kan', 'male', 'I am Superman!', ['Tony, Alice']);
+const woman = new Human('Kandy', 'female', 'Who is there?', ['Kan, Andy']);
+const catWoman = new Human('Lili', 'female', 'meow-meow! Who is there?', ['Tor, Iron-man']);
+
+let inhabitants = [dog, cat, woman, man, catWoman];
+
+inhabitants.forEach(element => {
+  print(element.getInfo());
+});
```
OleksiyRudenko: A bonus learning:

If you rename method `getInfo` into `toString` in all your classes, you will have a magic method.

`.toString()` gets called implicitly in a string context.

So your code 
```
inhabitants.forEach(element => {
  print(element.getInfo());
});
```
would be
```
inhabitants.forEach(element => {
  print(element.toString());
});
```
which can be effectively shortened as
```
inhabitants.forEach(element => {
  print(element);
});
```

Arrow functions that have a single statement in their body do not require curly brackets.
This gives us
```
inhabitants.forEach(element => print(element));
```

As we do not transform `element` before passing it to `print` and also do not take a property of element (e.g. `element.prop` or `element.methodName()`) to pass it to print, and function `print` itself matches callback signature requirements, the code can be as short as
```
inhabitants.forEach(print);
```

`forEach` will call `print` with every next item from the array.

PR #275, opened 2019-01-05
memory-pair-game

OlgaFrontend:
### Memory Pair Game

[Demo](https://olgafrontend.github.io/kottans-frontend/task_memory-game/) |  [Code](https://github.com/OlgaFrontend/kottans-frontend/tree/master/task_memory-game)

Please, review.


submissions/OlgaDubynka/memory-pair-game/main.js
```diff
@@ -0,0 +1,72 @@
+const mainBox = document.querySelector('.game-board');
+const allCards = document.querySelectorAll('.game-card');
+let isFlippedCard = false;
+let blockBoard = false;
```
yevhenorlov: It's better to use consistent naming convention for all predicates. One of the more popular ones — prefix "is". 
`isFlippedCard`, `isBlockedBoard`

submissions/OlgaDubynka/memory-pair-game/main.js
```diff
@@ -0,0 +1,72 @@
+const mainBox = document.querySelector('.game-board');
+const allCards = document.querySelectorAll('.game-card');
+let isFlippedCard = false;
+let blockBoard = false;
+let firstCard, secondCard;
+
+allCards.forEach(card => card.addEventListener('click', flipCard));
```
yevhenorlov: This solution is not ideal — it creates many event listeners for basically the same logic, which will impact browser performance in large scale applications. This is a prefect opportunity for you to try "event delegation" pattern:

https://learn.javascript.ru/event-delegation

https://medium.com/@bretdoucette/part-4-what-is-event-delegation-in-javascript-f5c8c0de2983

I understand that I ask you to change a big part of your application, but this would be a great training exercise. Feel free to ask for help if you need it.

submissions/OlgaDubynka/memory-pair-game/main.js
```diff
@@ -0,0 +1,72 @@
...
+const checkCardsMatching = () => {
+  if (firstCard.dataset.img === secondCard.dataset.img) {
+    return blockCards();
+  }
+  return stopFlipCards();
+}
+
+const blockCards = () => {
+  firstCard.removeEventListener('click', flipCard);
+  secondCard.removeEventListener('click', flipCard);
```
yevhenorlov: I see now why you need separate event listeners ;) 
Still, you can do this via a class (`.solved`, for example) and ignore clicks if element contains that class.

submissions/OlgaDubynka/memory-pair-game/main.js
```diff
@@ -0,0 +1,72 @@
...
+
+const stopFlipCards = () => {
+  blockBoard = true;
+
+  setTimeout(() => {
+    firstCard.classList.remove('flip');
+    secondCard.classList.remove('flip');
+
+    resetSettings();
+  }, 1500);
```
yevhenorlov: I think that maybe 1500 is a bit too long, I felt like the game was blocking me from playing.

submissions/OlgaDubynka/memory-pair-game/main.js
```diff
@@ -0,0 +1,72 @@
...
+
+const resetSettings = () => {
+  [isFlippedCard, blockBoard] = [false, false];
+  [firstCard, secondCard] = [null, null];
+}
+
+(function shuffleCards() {
+  allCards.forEach(item => {
+    let randomItemPos = Math.floor(Math.random() * 16);
+    item.style.order = randomItemPos;
```
yevhenorlov: cute solution :)

submissions/OlgaDubynka/memory-pair-game/main.js
```diff
@@ -0,0 +1,72 @@
...
+
+
+
+
+
+
+
+
+
+
```
yevhenorlov: one empty line in the end is plenty, please remove the rest

submissions/OlgaDubynka/memory-pair-game/main.js
```diff
@@ -61,12 +56,8 @@ const resetSettings = () => {
...
-
-
-
-
-
-
-
+gameBoard.addEventListener('click', function(e) {
+  let targetCard = e.target.parentElement;
+  targetCard.classList.contains('game-card') && flipCard(targetCard);
```
yevhenorlov: this is better, but the style is not very declarative - you're selecting some abstract `parentElement` and then checking if it's the card you need. There's a simpler solution — you can directly select the closest parent that is `.game-card` like so:
```js
let targetCard = e.target.closest('.game-card');
flipCard(targetCard);
```

however, make sure you add one more validation case for `flipCard` input if `targetCard` is null (happens when you click between cards).

yevhenorlov: also, you can safely use `const targetCard` - this variable is limited to the scope of this function and is never reassigned.

submissions/OlgaDubynka/memory-pair-game/main.js
```diff
@@ -0,0 +1,66 @@
...
+    let randomItemPos = Math.floor(Math.random() * 16);
+    item.style.order = randomItemPos;
+  });
+})();
+
+gameBoard.addEventListener('click', function(e) {
+  const targetCard = e.target.closest('.game-card');
+  targetCard !== null &&
+  targetCard.classList.contains('game-card') && 
+  flipCard(targetCard);
```
yevhenorlov: Almost correct. 
This works, but imagine that later you have to call `flipCard` from some other function — now you have to do the same check again.
If you check for `targetCard !== null` inside the body of the `flipCard`, you don't have to do it every time you call it. It's a good practice to validate inputs inside the function that expects them, not outside of it.
You don't have to do the `targetCard.classList.contains('game-card')` check, it's redundant now.

OlgaFrontend:
I added last improvements) Thank you )

PR #280, opened 2019-01-05
create a tiny JS world OOP

AlexNugda:
### a tiny JS world OOP

[Demo]() |
[Code base](https://github.com/AlexNugda/frontend-2019-homeworks/blob/tiny-js-world-oop/submissions/AlexNugda/a-tiny-JS-world-oop/index.js)

Please, review.

### DIFF
--- a/submissions/AlexNugda/a-tiny-JS-world-oop/index.js
+++ b/submissions/AlexNugda/a-tiny-JS-world-oop/index.js
@@ -1,56 +1,55 @@
 class Inhabitant {
-  constructor(name, gender) {
+  constructor(name, gender, species, saying) {
     this.name = name;
     this.gender = gender;
+	this.species = species;
+	this.saying = saying || `Hello from ${this.name}`;
+  }
+  showInhabitant(){
+	  return `<strong>${this.species}</strong>, ${this.name}, ${this.gender}, saying: ${this.saying}, `;
   }
 }
 
 class Human extends Inhabitant {
-	constructor(name, gender){
-		super(name, gender);
-		this.species = "Human";
+	constructor(name, gender, species){
+		super(name, gender, species || "Human");
 		this.hands = 2;
 		this.legs = 2;
-		this.saying = `Hello from ${this.name}`;
+	}
+	showInhabitant(){
+		return super.showInhabitant() + `hands: ${this.hands}, legs: ${this.legs}`;
 	}
 }
 
 class Animal extends Inhabitant {
-	constructor(name, gender){
-		super(name, gender);
+	constructor(name, gender, species, saying){
+		super(name, gender, species, saying);
 		this.paws = 4;
+	}
+	showInhabitant(){
+		return super.showInhabitant() + `paws: ${this.paws}`;
 	}
 }
 
 class Cat extends Animal {
 	constructor(name, gender){
-		super(name, gender);
-		this.species = "Cat";
-		this.saying = "Meow";
+		super(name, gender, "Cat", "Meow");
 	}
 }
 
 class Dog extends Animal {
 	constructor(name, gender){
-		super(name, gender);
-		this.species = "Dog";
-		this.saying = "Woff";
+		super(name, gender, "Dog", "Woff");
 	}
 }
 
 class CatWoman extends Human {
 	constructor(name, gender){
-		super(name, gender);
-		this.species = "CatWoman";
+		super(name, gender, "CatWoman");
 		this.saying = new Cat().saying;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


submissions/AlexNugda/a-tiny-JS-world-oop/index.js
```diff
@@ -0,0 +1,56 @@
+class Inhabitant {
+  constructor(name, gender) {
+    this.name = name;
+    this.gender = gender;
+  }
```
OleksiyRudenko: Name and gender are not the only properties all inhabitants share.
Generalization is one of the key principles in OOP.

submissions/AlexNugda/a-tiny-JS-world-oop/index.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+class Human extends Inhabitant {
+	constructor(name, gender){
+		super(name, gender);
+		this.species = "Human";
```
OleksiyRudenko: `super` is a regular function (almost), which resolves into base class constructor call.
I.e. here `super()` resolves into `Inhabitant.constructor()`.
You know that a function called from another function may have different number of arguments, order of arguments and some may be literals like strings, numbers etc

submissions/AlexNugda/a-tiny-JS-world-oop/index.js
```diff
@@ -0,0 +1,56 @@
...
+		this.saying = new Cat().saying;
+	}
+}
+
+let inhabitants = [new Human("Alex", "male"), new Human("Jenny", "female"), new Cat("Murka", "female"), new Dog("Rex", "male"), new CatWoman("LadyCat", "female")];
+
+const showInhabitant = ( inhabitant ) => {
+	let{species, name, gender, hands, legs, paws, saying} = inhabitant;
+	return `<strong>${species}</strong>, ${name}, ${gender}, hands: ${hands || "0"}, legs: ${legs || "0"}, paws: ${paws || "0"}, saying: ${saying}`;
+}
```
OleksiyRudenko: Outer world must not know the data model of a class.
This requirement is resolved by making this function a method of an appropriate class.

Also method has to manage only properties that are defined within class. That's how encapsulation principle is followed.

Luckily there is a method overloading, which works like this:
1. in a subclass define a method with a name == name of the method in super class
1. from within re-defined method just call parent's method (`super.methodName()`) and supplement it with info based on properties unique to a given child class.

This way you avoid constructions like `legs: ${legs || "0"}, paws: ${paws || "0"},` (imagine you add classes for even-toyed animals, animals like rhinos, kangaroos etc)

AlexNugda:
Thanks for the review!
I corrected the code according to your advice.

submissions/AlexNugda/a-tiny-JS-world-oop/index.js
```diff
@@ -0,0 +1,55 @@
+class Inhabitant {
+  constructor(name, gender, species, saying) {
+    this.name = name;
+    this.gender = gender;
+	this.species = species;
+	this.saying = saying || `Hello from ${this.name}`;
```
OleksiyRudenko: Please, fix indentation here and across the code.
We write code for other people, firtshand, and neatly formatted code is something making code easier to read.

submissions/AlexNugda/a-tiny-JS-world-oop/index.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+  showInhabitant(){
+	  return `<strong>${this.species}</strong>, ${this.name}, ${this.gender}, saying: ${this.saying}, `;
```
OleksiyRudenko: Please, employ `Array.join()` here.
Bonus: rename this method into `toString()` (here and in child classes). Explanation will follow.

submissions/AlexNugda/a-tiny-JS-world-oop/index.js
```diff
@@ -0,0 +1,55 @@
...
+class CatWoman extends Human {
+	constructor(name, gender){
+		super(name, gender, "CatWoman");
+		this.saying = new Cat().saying;
+	}
+}
+
+let inhabitants = [new Human("Alex", "male"), new Human("Jenny", "female"), new Cat("Murka", "female"), new Dog("Rex", "male"), new CatWoman("LadyCat", "female")];
+
+inhabitants.forEach(inhabitant => print(inhabitant.showInhabitant()));
```
OleksiyRudenko: `toString` is a magic method that is called implicitly in a string context.
So having it defined `print(obj)` has the same effect as `print(obj.toString)`

Furthermore, since we do not transform an array element before using it within the callback and do not use its property or whatever else, it appears that `print` is a sufficient callback for `forEach`. So `[].forEach(item => someFunc(item))` gets optimized as `[].forEach(someFunc)`

PR #286, opened 2019-01-05
add memory pair game task

frostwolm:
### Memory – Pair Game

[Demo](https://frostwolm.github.io/memory-pair-game/) |
[Code base](https://github.com/frostwolm/memory-pair-game)

Please, review.


submissions/frostwolm/pair-game/js/app.js
```diff
@@ -0,0 +1,94 @@
+let clickedCards = [];
+let cardsNum = 12;
+let imgIDsArray = [];
+let _checkNow = false;
+
+init(cardsNum);
```
AMashoshyna: Function "init" has no formal parameters

submissions/frostwolm/pair-game/js/app.js
```diff
@@ -0,0 +1,94 @@
...
+
+function init(){
+    document.querySelector('.game-container').innerHTML = '';
```
AMashoshyna: You make this query several times. DOM queries are expensive in terms of performance and should not be repeated if the element is not expected to change. Please save reference to DOM node in a variable and refer to it.

submissions/frostwolm/pair-game/js/app.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+  if (clickedCards.length == 2) {
+    _checkNow = true;
+    setTimeout(compareCards, 500);
+  }
+}
+
+function compareCards() {
+  if (clickedCards[0].parentNode.parentNode.parentNode.dataset.imgid ==
```
AMashoshyna: Relying in the nesting order of HTML elements is very fragile: if your layout changes due to redesign, your code will break. Please find a way to have controlled identifiers (data* attributes, ids, class names).

AMashoshyna:
Very nice job! 
Some bugs I found:
- when the cards do not match, the turn with different speed
- the game does not restart. If I have won and start a new game without reloading the page, the game alerts "Win" every time a turn new card pair

image


As for code - please see my comments

AMashoshyna:
Sorry for taking so long with the answer - I had to do some research.
So, the best solution I could come up with is to move `setTimeout` inside `forEach` callback like this:
```javascript
    clickedCards.forEach(cardElem => {
      setTimeout(() => flip(cardElem), 100);
    })
```
Pay attention to arrow function inside `setTimeout`: callback passed to `setTimeout` should be a function, not an expression.
I will ask other mentors for better ideas and let you know if we find something.

frostwolm:
Done

AMashoshyna:
Well done!

PR #290, opened 2019-01-05
Friends App

AlinaLadybug:
### Friends App 

[Demo](https://alinaladybug.github.io/friends-app/) |
[Code base](https://github.com/AlinaLadybug/friends-app)

Please, review.


submissions/AlinaLadybug/friends-app/js/script.js
```diff
@@ -0,0 +1,204 @@
+
+const URL = 'https://randomuser.me/api/?results=50';
+let personCards;
+let usersData = [];
+let filteredData = [];
+
+document.addEventListener("DOMContentLoaded", ready);
+function ready() {
```
AMashoshyna: Let's find a more suitable name for this function. It should better be a verb.

submissions/AlinaLadybug/friends-app/js/script.js
```diff
@@ -0,0 +1,204 @@
...
+function createThumb(info) {
+    let thumb = document.createElement('div');
+    thumb.classList.add('thumb');
+    let picture = info['picture'];
+    thumb.style.backgroundImage = `url(${picture['large']})`;
+    return thumb;
+}
+
+function getInfo() {
+    personCards = document.querySelectorAll('.card-body');
```
AMashoshyna: looks more like `setInfo`, because functions called get<something> usually should return something and are called to receive function value.

submissions/AlinaLadybug/friends-app/js/script.js
```diff
@@ -0,0 +1,204 @@
...
+function createContent(results, getInfo) {
+    results.forEach(result => {
+        createCard(result);
+    });
+    getInfo();
+}
+
+
+function createCard(info) {
+    let container = document.querySelector('.container .row');
```
AMashoshyna: This query will be made for every item in the list, but this element probably won't change. It is better to save it in a variable outside of loop.

submissions/AlinaLadybug/friends-app/js/script.js
```diff
@@ -0,0 +1,204 @@
...
+    let radios = document.querySelectorAll('.radio-inline input');
+    radios.forEach(radio => {
+        radio.addEventListener('click', sortbygender);
+    })
+}
+
+
+function sort({ target }) {
+    let className = target.classList[0];
+    let arrows = document.querySelectorAll('.nav-item img')
```
AMashoshyna: This logic does not belong to sorting. Let's put it inside a separate function

submissions/AlinaLadybug/friends-app/js/script.js
```diff
@@ -0,0 +1,204 @@
...
+                if (a['name']['first'] > b['name']['first'])
+                    return 1;
+            })
+            break;
+    }
+
+    createContent(filteredData, getInfo);
+}
+
+function sortbygender({ target }) {
```
AMashoshyna: Variable names containing multiple words should be in camelCase

AMashoshyna:
Hi @AlinaLadybug! Great job!

UI & UX:
- please add possibility to reset the filters. For example, if I choose a gender, there is no possibility to "un-choose" it. The input stays checked event after I choose other sorting and it is no linger valid.
- pushing "Enter" with any value in "Search" filed causes the page to reload (hint: might have something to do with default html form element behavior)

Code: please see my comments

AlinaLadybug:
@AMashoshyna Thanks a lot for your comments!) Please check changes.

submissions/AlinaLadybug/friends-app/js/script.js
```diff
@@ -23,19 +25,37 @@ function addListeners() {
...
     arrows.forEach(arrow => {
         arrow.classList.remove('active-arrow');
-    })
+    });
+    clearContent();
+    createContent(usersData, setInfo);
+}
+
+function sort({ target }) {
+    let className = target.classList[0];
```
AMashoshyna: I would rather use data* attribute to store data here. Reason: you rely not only on css, but specifically on classes order, which may be subject to change. It is better to keep presentation and logic separate.

submissions/AlinaLadybug/friends-app/js/script.js
```diff
@@ -92,6 +112,8 @@ function clearContent() {
     let row = document.querySelector('.row');
```
AMashoshyna: `row` is slightly misleading name, because usually we call only one line of items a row. Here we see multiple lines of items. I'd say it's rather a grid, than a row. But it's up to you, just a suggestion.

AMashoshyna:
Great job! I have some more questions:
1. You import jQuery library. What do you use it for?
2. Does your app support mobile view?

Also please see my comments.

AlinaLadybug:
@AMashoshyna
1. I use bootstrap's navbar and its collapsing system, soit requires jQuery for normal work.
2. Now app supports mobile view)

AlinaLadybug:
@AMashoshyna Sorry, so what?)

AMashoshyna:
Good job! Sorry for late reply.

PR #293, opened 2019-01-05
Friends App

dnzln:
### Friends App

[Demo](https://dnzln.github.io/friends-app/) |
[Code base](https://github.com/dnzln/friends-app)

Please, review.

### DIFF
--- a/submissions/dnzln/friends-app/app.js
+++ b/submissions/dnzln/friends-app/app.js
@@ -1,0 +1,105 @@
+const MAIN_CONTAINER = document.getElementById('main-content');
+const MAIL_INPUT = document.querySelector('input[value=male]');
+const FEMAIL_INPUT = document.querySelector('input[value=female]');
+const FILTERS = document.querySelector('.filters');
+const SEARCH_FIELD = document.getElementById('search-field');
+const LABELS_IMG = document.getElementsByClassName('radio-img');
+var usersArray = [];
+
+fetch('https://randomuser.me/api/?results=48')
+    .then( response => {response.json()
+    .then( data => {
+        usersArray = data.results
+        printingUsers(usersArray)
+	})
+    });
+
+function printingUsers(usersArray) {
+    isChecked(); 
+    MAIN_CONTAINER.innerHTML = '';
+    let fragment = document.createDocumentFragment();
+    usersArray.forEach(
+        function(user) {
+
+            if(FEMAIL_INPUT.checked) if(user.gender == 'female') return;
+            if(MAIL_INPUT.checked) if(user.gender == 'male') return;
+            if(!`${user.name.first} ${user.name.last}`.includes(SEARCH_FIELD.value)) return;
+            
+
+            let block = document.createElement('div');
+            block.classList.add('blocks');
+            fragment.appendChild(block);
+
+            let card = document.createElement('div');
+            card.classList.add('user-card');
+            block.appendChild(card);
+            
+            let userImage = document.createElement('img');
+            userImage.classList.add('user-pic');
+            userImage.setAttribute('src', user.picture.large);
+            card.appendChild(userImage);
+            
+            let userName = document.createElement('p');
+            userName.classList.add('user-name');
+            userName.innerHTML = `${user.name.first} ${user.name.last}`;
+            card.appendChild(userName);
+
+            let userData = document.createElement('p');
+            userData.classList.add('user-data');
+            userData.innerHTML = `<strong>${user.dob.age}</strong>, from ${user.location.city}`;
+            card.appendChild(userData);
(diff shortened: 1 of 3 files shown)


submissions/dnzln/friends-app/app.js
```diff
@@ -0,0 +1,105 @@
+const MAIN_CONTAINER = document.getElementById('main-content');
+const MAIL_INPUT = document.querySelector('input[value=male]');
+const FEMAIL_INPUT = document.querySelector('input[value=female]');
+const FILTERS = document.querySelector('.filters');
+const SEARCH_FIELD = document.getElementById('search-field');
+const LABELS_IMG = document.getElementsByClassName('radio-img');
+var usersArray = [];
```
AMashoshyna: You can use `const` here too. Vars introduce unpredictability into the code

submissions/dnzln/friends-app/app.js
```diff
@@ -0,0 +1,105 @@
...
+
+fetch('https://randomuser.me/api/?results=48')
+    .then( response => {response.json()
+    .then( data => {
+        usersArray = data.results
+        printingUsers(usersArray)
+	})
+    });
+
```
AMashoshyna: Nesting of promises is possible, but here's a better way (attention to how Promise chain is built):
```javascript
fetch('https://randomuser.me/api/?results=48')
    .then( response => response.json())
    .then( data => {
        usersArray = data.results
        printingUsers(usersArray)
	})
 ```

File: submissions/dnzln/friends-app/app.js [hunk 3]
```diff
@@ -0,0 +1,105 @@
...
+function printingUsers(usersArray) {
+    isChecked(); 
+    MAIN_CONTAINER.innerHTML = '';
+    let fragment = document.createDocumentFragment();
+    usersArray.forEach(
+        function(user) {
+
+            if(FEMAIL_INPUT.checked) if(user.gender == 'female') return;
```
AMashoshyna: I suggest that you do the sorting by gender and search phrase within the switch statement. This logic does not belong in `printUsers` function - this function should receive ready array to be displayed.

submissions/dnzln/friends-app/app.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+    }
+);
+
+SEARCH_FIELD.addEventListener('input', function() {
+    printingUsers(usersArray); 
+});
+
+function isChecked(){
+    for(let i = 0; i < LABELS_IMG.length; i++) {
```
AMashoshyna: Do I understand correctly, that this function is there for styling purposes? If so, please consider to use `:checked` pseudo-class selector instead of adding "checked" class and move this logic to CSS completely.

AMashoshyna:
Hi @dnzln!

Here are general impressions about UI & UX:

- Longer location names must be handled in layout:
image

- It is good to indicate clickable items by changing the cursor appearance or similar.

- Suggestion for improvement (optional): when: also when an item is active (for example, users are sorted by age) it should be indicated and the corresponding button should be disabled.

- Suggestion for improvement (optional): when you show results of search, add the number of users found. When no user matches the search phrase, show message "no matches" or similar. Empty screen is confusing.

Code: please see my comments.

submissions/dnzln/friends-app/app.js
dnzln: Tried, it's return error: Uncaught (in promise) TypeError: Assignment to constant variable.
    at fetch.then.then.data

dnzln: In that case I got error: Uncaught (in promise) TypeError: Assignment to constant variable.
    at fetch.then.then.data

dnzln:
Done. Tried to implement all of your suggestions.

submissions/dnzln/friends-app/app.js
AMashoshyna: Yes, my bad. I missed that this value is overwritten later.

dnzln:
@AMashoshyna
Tell me, please, is there anything else that I have to fix in this task, or it's can be merged?

AMashoshyna:
Hi @dnzln! You should trim the user input before making the filtering/sorting. This should be specified by UX designer, but most often if you input empty space, no action should be made.

AMashoshyna:
@dnzln Yes, your should not search for empty line (either it is one space or multiple). Easiest way to achieve this is to use [String.prototype.trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim) method on user input.

dnzln:
Done. Please check if everything is OK now.

AMashoshyna:
Perfect!

PR #297, opened 2019-01-06
OOP Exercise

ivarshavets:
### OOP Exercise

[Demo](https://ivarshavets.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/ivarshavets/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/ivarshavets/js-post-oop/index.js
+++ b/submissions/ivarshavets/js-post-oop/index.js
@@ -6,81 +6,61 @@
 
 // ======== OBJECTS DEFINITIONS ========
 class Inhabitant {
-  constructor(name, gender, speak, friends = null){
+  constructor(species, name, gender, phrase, friends = null, legs = 2){
+    this.species = species
     this.name = name
     this.gender = gender
-    this.speak = speak
+    this.phrase = phrase
     this.friends = friends
+    this.legs = legs
   }
 
   toString() {
-    let info = [this.name, this.gender, this.speak].join("; ")
-    if ( this.friends ) {
-      info += `; ${this.friends.join(", ")}`
-    }
-    return info
-  }
-  print () {
-    print( this.toString() )
+    return [this.species, this.name, this.gender, this.phrase, this.friends, this.legs].join("; ")
   }
 }
 
 class Animal extends Inhabitant {
-  constructor(name, gender, speak, friends, species) {
-    super(name, gender, speak, friends)
-    this.species = species
-    this.legs = 4
-  }
-
-  toString() {
-    return `${this.species}; ${this.legs} legs; ${super.toString()}`
+  constructor(species, name, gender, phrase, friends, legs) {
+    super(species, name, gender, phrase, friends, 4)
   }
 }
 
 class Dog extends Animal {
   constructor(name, gender, friends) {
-    super(name, gender, "woof-woof", friends)
-    this.species = "dog"
+    super("dog", name, gender, "woof-woof", friends)
   }
 }
 
 class Cat extends Animal {
   constructor(name, gender, friends) {
-    super(name, gender, "meow", friends)
-    this.species = "cat"
+    super("cat", name, gender, "meow", friends)
   }
 }
 
 class Human extends Inhabitant {
-  constructor(name, gender, speak, friends) {
-    super(name, gender, speak, friends)
-    this.species = "human"
-    this.legs = 2
+  constructor(name, gender, phrase, friends) {
+    super("human", name, gender, phrase, friends, 2)
     this.hands = 2
   }
 
   toString() {
-    return `${this.species}; ${this.legs} legs; ${this.hands} hands; ${super.toString()}`
+    return [super.toString(), this.hands].join("; ")
   }
 }
 
 class Catwoman extends Human {
   constructor(name, friends) {
-    super(name, "female", friends)
-    this.species = "cat-woman"
-    this.speak = new Cat().speak
+    super("cat-woman", name, "female", friends)
   }
 }
 
-const inhabitants = [
-  new Dog("Ghost", "male", ["Jon Snow", "Stark"]),
-  new Cat("Misty", "female"),
-  new Human("Ygritte", "female", "Hi!", ["Jon Snow"]),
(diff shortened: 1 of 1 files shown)


submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,87 @@
...
+  print () {
+    print( this.toString() )
+  }
+}
+
+class Animal extends Inhabitant {
+  constructor(name, gender, speak, friends, species) {
+    super(name, gender, speak, friends)
+    this.species = species
+    this.legs = 4
```
OleksiyRudenko: All inhabitants have legs and attributed to a certain species. So these props should be a part of the basemost class.

submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+  toString() {
+    return `${this.species}; ${this.legs} legs; ${super.toString()}`
+  }
+}
+
+class Dog extends Animal {
+  constructor(name, gender, friends) {
+    super(name, gender, "woof-woof", friends)
+    this.species = "dog"
```
OleksiyRudenko: If super had `species` param you could pass `"dog"` as a literal.

submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+
+class Human extends Inhabitant {
+  constructor(name, gender, speak, friends) {
+    super(name, gender, speak, friends)
+    this.species = "human"
+    this.legs = 2
+    this.hands = 2
+  }
```
OleksiyRudenko: The only unique prop for humans is hands.

submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+  toString() {
+    return `${this.species}; ${this.legs} legs; ${this.hands} hands; ${super.toString()}`
+  }
```
OleksiyRudenko: There is still sense to employ `Array.join`, However, once you have species and legs properties of the basemost class this might not be the case anymore.

submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+class Catwoman extends Human {
+  constructor(name, friends) {
+    super(name, "female", friends)
+    this.species = "cat-woman"
+    this.speak = new Cat().speak
```
OleksiyRudenko: Not a really good idea to borrow any property for a class from a class instance.
es6 doesn't offer an elegant syntax for composition, so let's leave this as it is.

Alternatively you may try finding a different way. You might borrow something from class `__proto__` if you had methods to set and get properties.

submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,87 @@
...
+  new Dog("Ghost", "male", ["Jon Snow", "Stark"]),
+  new Cat("Misty", "female"),
+  new Human("Ygritte", "female", "Hi!", ["Jon Snow"]),
+  new Human("Jon Snow", "male", "Hey yo!", ["Ygritte", "Ghost"]),
+  new Catwoman("Selina")
+]
+
+// ======== OUTPUT ========
+inhabitants.forEach(inhabitant => {
+  inhabitant.print()
```
OleksiyRudenko: Please, replace with `print(inhabitant.toString())`.

BTW, you may omit `.toString()` as it is a magic method and gets called implicitly in a string context. So, `print(inhabitant)` would work.

Even further, when you do not need to transform an argument to a callback or take e.g. a property from it there is another improvement possible:

`[].forEach(item => someFunction(item))` effectively becomes `[].forEach(someFunction)` as
`someFunction` is already a function you can use as a callback.

submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,87 @@
...
+  console.log(inhabitant)
```
OleksiyRudenko: `print` does `console.log` for you.

submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,87 @@
...
+   Code repository: https://github.com/ivarshavets/a-tiny-JS-world
+   Web app: https://ivarshavets.github.io/a-tiny-JS-world/
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+class Inhabitant {
+  constructor(name, gender, speak, friends = null){
+    this.name = name
+    this.gender = gender
+    this.speak = speak
```
OleksiyRudenko: By convention simple properties names should start with a noun.
I would suggest `phrase` or anything alike.

submissions/ivarshavets/js-post-oop/index.js
ivarshavets: Not sure yet how to implement it with `__proto__`

Could it be an option - passing `cat.phrase` as an argument for `catwoman`?

OleksiyRudenko: @ivarshavets Let it be so for now.
It's an optional task. We'll have a chance to dig deeper into it during the offline course phase.

submissions/ivarshavets/js-post-oop/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+const dog = new Dog("Ghost", "male", ["Jon Snow", "Stark"])
+const cat = new Cat("Misty", "female")
+const woman = new Human("Ygritte", "female", "Hi!", ["Jon Snow"])
+const man = new Human("Jon Snow", "male", "Hey yo!", ["Ygritte", "Ghost"])
+const catwoman = new Catwoman("Selina", cat.phrase)
+const inhabitants = [dog, cat, woman, man, catwoman]
+
+// ======== OUTPUT ========
+inhabitants.forEach(inhabitant => print(inhabitant))
```
OleksiyRudenko: `inhabitants.forEach(print)` is an equivalent as `print` signature per se fits `.forEach` callback function signature requirement.

PR #298, opened 2019-01-06
add js_dom task

iamsokol:
### task js_dom

[Demo](https://iamsokol.github.io/js_dom/)   
[Code base](https://github.com/iamsokol/iamsokol.github.io/tree/master/js_dom)

Please, review.

### DIFF
--- a/submissions/iamsokol/js_dom/index.html
+++ b/submissions/iamsokol/js_dom/index.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>js_dom</title>
+        <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro" rel="stylesheet">
+        <link rel="stylesheet" href="style.css" />
+    </head>
+    <body>
+        <div id="menu" class="menu"></div>
+        <div id="main" class="main"></div>
+        <script src="script.js"></script>
+    </body>
+</html>
--- a/submissions/iamsokol/js_dom/script.js
+++ b/submissions/iamsokol/js_dom/script.js
@@ -26,37 +26,24 @@ const catsArray = [
     }
 ]
 
-let menuBlock = document.createElement('div');
 let renderCats = (position) => {
-    new Cat(catsArray, position);
+    let item = catsArray[position];
+    let catItem = `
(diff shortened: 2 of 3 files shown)


AMashoshyna:
Hi @iamsokol! Nice job!

Please fix the following issues:
- PR formatting: please fix markdown in the first comment (links are not displayed correctly):
image
- names of classes should start with uppercase letter and be nouns
- let's split instantiation of an object and object's operations. For example, `renderMenu` class (unfortunate naming - see previous point). When an instance of this class is created, during creation this instance reaches to HTML element `block` via global scope and changes it. This is a rather unexpected behavior. It would be better if class constructor would not do any mutating operations. Instead you should call renderMenu.renderItem method explicitly. Also, reference to `block` element should be passed as parameter to renderItem method. In short: creating menu and rendering should be two different things.
- eventListeners: please use event delegation instead of adding event listeners to all list items. Add one event listener to container element instead.

Let me know if you need help.

iamsokol:
@AMashoshyna, I fixed everything

AMashoshyna:
Ok, I see two major problems with your code:
- global variables
- incorrect use of OOP features

Let me explain

It is great, that you have used the knowledge gained in the OOP-related tasks. But usually when a class instance is created, it is created to be used further. In your case you don't use the instances of **Cat** class you create. Basically you only use the side-effect (DOM manipulation provided by `renderPosition` and `renderItem` methods). This is rather unexpected code behavior and a person reading your code would need to read whole classes to understand what's going on.

Concerning global variables: you classes reach out to `menu` and `menuBlock` variables in the global scope. Excessive use of global variables is often considered source of many bugs.

My suggestion for both of the problems is to split the two things: instantiation of class instance and rendering. This suggestion is inspired by React component API:
1. Remove rendering method call from class constructors
2. Call render method (`renderPosition`, `renderItem`) explicitly 
3. Pass references to DOM elements explicitly to rendering methods

Let me know if you have question or need more explanations.

Concerning your question: you are right, `menuBlock` is not in the DOM when catsArray is iterated.

iamsokol:
Could you please look my code again. I made some changes

AMashoshyna:
Nice job! I like your solution with simple functions instead of classes even better!

AMashoshyna:
@iamsokol  One important thing that I missed during review, but I'd like you to notice: you refer to DOM elements by id directly (menu, main). This works, because global variables are created for elements with ids. However I would not recommend to use this approach, because is it not obvious what's happening. Always better to query elements explicitly (with document.getElementById, document.querySelector etc.).

PR #301, opened 2019-01-07
js oop practice

iamsokol:
### js oop practice

[Demo](https://iamsokol.github.io/a-tiny-JS-world-2/)  
[Code base](https://github.com/iamsokol/iamsokol.github.io/tree/master/a-tiny-JS-world-2)

Please, review.

### DIFF
--- a/submissions/iamsokol/tiny-js-world-2/index.js
+++ b/submissions/iamsokol/tiny-js-world-2/index.js
@@ -11,9 +11,9 @@ class Inhabitant {
     }
 }
 
-class Man extends Inhabitant {
-    constructor(name, gender, saying, legs = 2, hands = 2) {
-        super('human', name, gender, saying, legs);
+class Human extends Inhabitant {
+    constructor(species, name, gender, saying, legs = 2, hands = 2) {
+        super(species, name, gender, saying, legs);
         this.hands = hands;
     }
     toString(){
@@ -21,13 +21,15 @@ class Man extends Inhabitant {
     }
 }
 
-class Woman extends Inhabitant {
-    constructor(name, gender, saying, legs = 2, hands = 2) {
-        super('human', name, gender, saying, legs);
-        this.hands = hands;
+class Man extends Human {
+    constructor(name, saying) {
+        super('human', name, 'male', saying);
     }
-    toString(){
-        return super.toString() + this.hands + ';';
+}
+
+class Woman extends Human {
+    constructor(name, saying) {
+        super('human', name, 'female', saying);
     }
 }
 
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@iamsokol Thank you for your submission.
Does this PR relate to **Build a Tiny JS World** or **OOP Exercise** task?

iamsokol:
I have already done Build a Tiny JS World task.
This PR is for OOP exercise

OleksiyRudenko:
@iamsokol Thank you for the explanation. Please, fix the task name, as "2" in task name is ambiguous and might mean the second attempt.

In many big companies it is important to identify and refer properly an issue (task, bug etc) your commit/PR resolves. It's a part of the communication flow and responsibility of contributors.

submissions/iamsokol/tiny-js-world-2/index.js
```diff
@@ -0,0 +1,62 @@
+class Inhabitant {
+    constructor(species, name, gender, saying, legs = 4) {
+        this.species = species;
+        this.name = name;
+        this.gender = gender;
+        this.saying = saying;
+        this.legs = legs;
+    }
+    toString(){
+        return [this.species, this.name, this.gender, this.saying, this.legs + '; '].join ('; ');
```
OleksiyRudenko: Not critical but adding final `;` in the end of string construction would be more obvious.

submissions/iamsokol/tiny-js-world-2/index.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+class Man extends Inhabitant {
+    constructor(name, gender, saying, legs = 2, hands = 2) {
```
OleksiyRudenko: Do we need to ask a user for a gender when instantiating `Man`?

Another point.

Inhabitant gets extended by `Cat`, `Dog`, `Man`, `Woman`.
Cats and dogs are species of their own. Men and women aren't species.
As these classes are on the same level of class hierarchy some consistency among those is required. E.g. either define classes `HeDog`, `SheDog`, `HeCat`, `SheCat` or generalize `Man` and `Woman` into a species.

Please, make class hierarchy consistent.

submissions/iamsokol/tiny-js-world-2/index.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+class Woman extends Inhabitant {
+    constructor(name, gender, saying, legs = 2, hands = 2) {
+        super('human', name, gender, saying, legs);
+        this.hands = hands;
+    }
+    toString(){
+        return super.toString() + this.hands + ';';
+    }
```
OleksiyRudenko: There is much repetition between `Man` and `Woman`. One of the points of OOP is to avoid repetition keeping the code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

submissions/iamsokol/tiny-js-world-2/index.js
iamsokol: What do you mean? I have ' ; ' in this string.

OleksiyRudenko: Yes, you have.
But it reads like "let's join info on species, name, gender, saying and legs combined with ';' for whatever reason".  You will have same number of characters in your code if you move `+ ';'` to the end of the construction, but it will read "let's join ... and legs with a ';' and yet another ';' to close the string".

We write code for other people. firsthand.

BTW, how about taking a learning from OO JS task? I mean `const`.

PR #303, opened 2019-01-07
OOP exercise.

krash1408:
### OOP exercise.

[Demo](https://krash1408.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/krash1408/a-tiny-JS-world/tree/oop)

Please, review.

### DIFF
--- a/submissions/krash1408/tiny-js-world-OOP/index.js
+++ b/submissions/krash1408/tiny-js-world-OOP/index.js
@@ -7,33 +7,38 @@
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
 class Create {
-    constructor(gender, name, age, say, species, legs, hands) {
+    constructor(gender, name, age, say, species, legs) {
         this.gender = gender;
         this.name = name;
         this.say = say;
         this.species = species;
         this.age = age;
         this.legs = legs;
-        this.hands = hands;
     }
     whatReturn() {
-        return print(`Gender: ${this.gender}, name: ${this.name}, say: ${this.say}, species: ${this.species}, age: ${this.age}, legs ${this.legs}, hands: ${this.hands}`);
+     if (this.hands == undefined) {
+        return (`Gender: ${this.gender}, name: ${this.name}, say: ${this.say}, species: ${this.species}, age: ${this.age}, legs: ${this.legs}`);
+     } else {
+         return (`Gender: ${this.gender}, name: ${this.name}, say: ${this.say}, species: ${this.species}, age: ${this.age}, legs: ${this.legs}, hands: ${this.hands}`)
+     }
     }
 }
 
 class Human extends Create {
     constructor ( gender, name, age, say, species = 'Human', legs = '2', hands = '2') {
-        super (gender, name, age, say, species, legs, hands);
+        super (gender, name, age, say, species, legs);
+        this.hands = hands;
     }
+    
 }
 class Dog extends Create {
-    constructor (gender, name, age, say = 'Woof!', species = 'Dog', legs = '4', hands = '0') {
-        super (gender, name, age, say, species, legs, hands);
+    constructor (gender, name, age, say = 'Woof!', species = 'Dog', legs = '4') {
+        super (gender, name, age, say, species, legs);
     }
 }
 class Cat extends Create {
-    constructor (gender, name, age, say = 'Meow!', species = 'Cat', legs = '4', hands = '0') {
-        super (gender, name, age, say, species, legs, hands);
+    constructor (gender, name, age, say = 'Meow!', species = 'Cat', legs = '4') {
+        super (gender, name, age, say, species, legs);
     }
 }
 
@@ -60,6 +65,4 @@ let dog = new Dog('male', 'Scooby', 7);
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
 */
-[woman, man, cat, dog].forEach(value => {
-    value.whatReturn();
-})
+[man, woman, cat, dog].forEach( val => print(val.whatReturn(), 'div') )
--- a/submissions/krash1408/tiny-js-world-OOP/index.js
+++ b/submissions/krash1408/tiny-js-world-OOP/index.js
@@ -16,20 +16,17 @@ class Create {
         this.legs = legs;
     }
     whatReturn() {
-     if (this.hands == undefined) {
         return (`Gender: ${this.gender}, name: ${this.name}, say: ${this.say}, species: ${this.species}, age: ${this.age}, legs: ${this.legs}`);
-     } else {
-         return (`Gender: ${this.gender}, name: ${this.name}, say: ${this.say}, species: ${this.species}, age: ${this.age}, legs: ${this.legs}, hands: ${this.hands}`)
-     }
     }
 }
-
 class Human extends Create {
     constructor ( gender, name, age, say, species = 'Human', legs = '2', hands = '2') {
         super (gender, name, age, say, species, legs);
         this.hands = hands;
     }
-    
+    whatReturn() {
+        return (super.whatReturn() + `, hands: ${this.hands}`)
+    }
 }
 class Dog extends Create {
     constructor (gender, name, age, say = 'Woof!', species = 'Dog', legs = '4') {
@@ -41,12 +38,10 @@ class Cat extends Create {
         super (gender, name, age, say, species, legs);
     }
 }
-
 let woman = new Human('male', 'Jane', 57, 'Holla!');
 let man = new Human('male' , 'John', 57, 'I love development.');
 let cat = new Cat('male', 'Simon', 3);
 let dog = new Dog('male', 'Scooby', 7);
-
 // ======== OUTPUT ========
 /* Use print(message) for output.
    Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
--- a/submissions/krash1408/tiny-js-world-OOP/index.js
+++ b/submissions/krash1408/tiny-js-world-OOP/index.js
@@ -1,0 +1,74 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/krash1408/a-tiny-JS-world.git
+   Web app: https://krash1408.github.io/a-tiny-JS-world/
+   */
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Create {
+    constructor(name, gender, inhabitant,  age, legs) {
+        this.name = name;
+        this.gender = gender;
+        this.inhabitant = inhabitant;
+        this.age = age;
+        this.legs = legs;
+        this.say = '';
+    }
+    whatSayAnimal() {
+        if(this.hands == undefined) {
+            this.say = 'I can\'t! I have paws!';
+        }
+    }
+    whatSayHuman() {
+        if (this.inhabitant == 'man') {
+            this.say = 'It`s show time!';
+        } else if (this.inhabitant == 'woman') {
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@krash1408 Thank you for your submission. Please, provide a demo.

krash1408:
Done. Demo's created.

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,74 @@
...
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+    [cat, dog].forEach(val => {
+        print(`Hi! My name is - ${val.name}. I\'m ${val.age} years old. But I\'m steel young to be a agile ${val.inhabitant}. I have ${val.legs} legs for this. Work? ${val.say}`);
+    });
```
OleksiyRudenko: Outer world wouldn't know anything about classes' internals. Outer world communicates with classes and instances via methods.
There might be a method in a base class that responds with (returns) a nicely formatted string representing props know to the  base class.
Child classes may define similar method where they call parent class method and supplement its result with information specific to that particular class.
Please, check **MDN reference**, **"Super" and "Extends"....**, and **ES6 classes cheat snippet** from materials in the [original repo](https://github.com/OleksiyRudenko/a-tiny-JS-world/blob/master/README.md#learn-on-your-own)

krash1408:
Not sure about the changes. Have problems with the translation from English.

OleksiyRudenko:
@krash1408 No worries. Let's do in smaller steps.

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,65 @@
...
+        this.gender = gender;
+        this.name = name;
+        this.say = say;
+        this.species = species;
+        this.age = age;
+        this.legs = legs;
+        this.hands = hands;
+    }
+    whatReturn() {
+        return print(`Gender: ${this.gender}, name: ${this.name}, say: ${this.say}, species: ${this.species}, age: ${this.age}, legs ${this.legs}, hands: ${this.hands}`);
```
OleksiyRudenko: `print` doesn't return anything (to be prices all functions that do not have `return` statement return `undefined`).
But we also do not want to print anything here as classes should be agnostic of media/UI (unless they are specifically designed to manage media/UI).
So, just return formatted string and print in the main job part of the code.

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,65 @@
...
+// Define your objects here
+class Create {
+    constructor(gender, name, age, say, species, legs, hands) {
+        this.gender = gender;
+        this.name = name;
+        this.say = say;
+        this.species = species;
+        this.age = age;
+        this.legs = legs;
+        this.hands = hands;
```
OleksiyRudenko: Not all species have hands by natural design. Imagine we need to extend our world with even-toed animals, and animals like rhinos and kangaroos. This way your base class gets cluttered very soon. OOP has a number of tools to manage this.

So, remove hands from base class, and add those to the appropriate class.

submissions/krash1408/tiny-js-world-OOP/index.js
OleksiyRudenko: As the base class doesn't own `hands` property it would be counter-OOP and counter-SOLID to use this property in any method of the base class.

There is a solution:
Every child class may override a method from the base class. I.e. a child class may define a method with a name exactly the same as in the base class. This child class method can call a method from the parent class (`super.methodName()`), supplement it with info specific to this child class and return the aggregated result.

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+class Human extends Create {
+    constructor ( gender, name, age, say, species = 'Human', legs = '2', hands = '2') {
+        super (gender, name, age, say, species, legs, hands);
+    }
+}
+class Dog extends Create {
+    constructor (gender, name, age, say = 'Woof!', species = 'Dog', legs = '4', hands = '0') {
+        super (gender, name, age, say, species, legs, hands);
```
OleksiyRudenko: We do not need to ask user for species or give a user an opportunity to specify a different class. Also dogs do not have hands.

Please, explain why do you need those?

Note that `super()` actually resolves to (in this case of `Dog`) `Create.constructor()`. So `super` is a regular function. As a function `super` may receive anything as arguments - variables, literals (strings, numbers, Boolean values, etc).

The number, structure and order of arguments in child class constructor and `super` do not have to be exactly the same. It is not a rare case when you call one function from inside another one and they have totally different arguments. Example: your own function `whatReturn` calls `print` and they have different arguments.

If you need clarification on the above, please, let me know.

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,65 @@
...
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+*/
+[woman, man, cat, dog].forEach(value => {
+    value.whatReturn();
```
OleksiyRudenko: As `.whatReturn()` will return a string you need then to pass the result from `.whatReturn()` to `print`.

submissions/krash1408/tiny-js-world-OOP/index.js
krash1408: Yes, please. It's hardest part for me.

OleksiyRudenko: @krash1408 What is the first part that is not clear?
I will also reformulate the question in paragraph 2: Please, explain why do you have `hands` among other `Dog.constructor` parameters?

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,68 @@
...
+        this.species = species;
+        this.age = age;
+        this.legs = legs;
+    }
+    whatReturn() {
+     if (this.hands == undefined) {
+        return (`Gender: ${this.gender}, name: ${this.name}, say: ${this.say}, species: ${this.species}, age: ${this.age}, legs: ${this.legs}`);
+     } else {
+         return (`Gender: ${this.gender}, name: ${this.name}, say: ${this.say}, species: ${this.species}, age: ${this.age}, legs: ${this.legs}, hands: ${this.hands}`)
+     }
```
OleksiyRudenko: > Every child class may override a method from the base class. I.e. a child class may define a method with a name exactly the same as in the base class. This child class method can call a method from the parent class (super.methodName()), supplement it with info specific to this child class and return the aggregated result.

Please, reconsider the quote from the past comment and address a particular part of it if anything is not clear enough.

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,68 @@
...
+    constructor ( gender, name, age, say, species = 'Human', legs = '2', hands = '2') {
+        super (gender, name, age, say, species, legs);
+        this.hands = hands;
+    }
+    
+}
+class Dog extends Create {
+    constructor (gender, name, age, say = 'Woof!', species = 'Dog', legs = '4') {
+        super (gender, name, age, say, species, legs);
+    }
```
OleksiyRudenko: > We do not need to ask user for species or give a user an opportunity to specify a different class.

When constructor has a parameter with a default value this means that there is an opportunity to specify a different species. We do not want this as this breaks OOP principles.

> Note that super() actually resolves to (in this case of Dog) `Create.constructor()`. So super is a regular function. As a function super may receive anything as arguments - variables, literals (strings, numbers, Boolean values, etc).

> The number, structure and order of arguments in child class constructor and super do not have to be exactly the same. It is not a rare case when you call one function from inside another one and they have totally different arguments. Example: your own function whatReturn calls print and they have different arguments.

The above means that you can call `super` with some literal arguments e.g. `super(variableA, variableB, 'string literal', 55); // 55 is a numeric literal ` - just like any other function.

Please, if unclear ask a question addressing a specific part of the comment above.

> What is the first part that is not clear?

Please, answer the questions. Your responses might clarify your way of thinking to a code reviewer, which in its turn helps to render more efficient support.

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,68 @@
...
+/* Print examples:
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+*/
+[man, woman, cat, dog].forEach( val => print(val.whatReturn(), 'div') )
```
OleksiyRudenko: Please, tune up your IDE so it ensures your code files end with an empty line. That's important as helps to avoid diffing a line of code where the only change is a change of a newline character.

OleksiyRudenko:
@krash1408 Please, resolve conversations you believe you have elaborated on. At the same time if anything is not clear as a question in a relevant conversation thread. If you're not going to implement any recommendation knowingly, please, give a reason in a relevant conversation thread. This is important communication flow, especially in learning. Otherwise it is very difficult to render appropriate help.

submissions/krash1408/tiny-js-world-OOP/index.js
krash1408: Not entirely clear where to explicitly specify the parameters in the child class or the parent? Since in the end all the same there is an appeal to the parent class.

OleksiyRudenko: Does the following explain?

> The above means that you can call super with some literal arguments e.g. super(variableA, variableB, 'string literal', 55); // 55 is a numeric literal - just like any other function.

OleksiyRudenko: > Since in the end all the same there is an appeal to the parent class.
You are right. But you do not have to ask a user about everything you need. You can supplement some default information.

E.g. when user does `new Dog`. A `Dog` constructor gets called and you know it is a `Dog`. So you can call a `super` specifying species as a literal (e.g. `"dog"`) without taking any from `Dog.constructor` params.

Does the above explain?

krash1408: I understood how to set global parameters through super. Buthow to set specific parameters, such as the _hands_ method. I described their basic values in the constructor, is this correct?

OleksiyRudenko: You do not need to propagate props specific to a certain class to the parent class.
That's the point of encapsulation.

submissions/krash1408/tiny-js-world-OOP/index.js
```diff
@@ -0,0 +1,63 @@
...
```
OleksiyRudenko: Your bonus learning:
If you rename `whatReturn` in all classes to `toString` you will have a magic method.
`.toString()` is called implicitly in string context. This means you can just `print(val)` (equivalent of `print(val.toString())`).
BTW, `print` second parameter is `'div'` so you do not need to specify it.
If you fix all things above you code can be as short as
`[man, woman, cat, dog].forEach(print)` for `print()` fits callback requirements.

PR #304, opened 2019-01-07
A tiny JS world (OOP)

vv2529:
### OOP exercise

[Demo](https://vv2529.github.io/a-tiny-JS-world) |
[Code base](https://github.com/vv2529/a-tiny-JS-world/blob/populate-world/index.js)

Please, review.

### DIFF
--- a/submissions/vv2529/a-tiny-js-world-oop/index.js
+++ b/submissions/vv2529/a-tiny-js-world-oop/index.js
@@ -7,13 +7,14 @@
 
 // ======== CLASS DEFINITIONS ========
 class Inhabitant {
-	constructor(type, name, gender, legs, hands, says){
+	constructor(type, name, gender, legs, hands, says, friends = []){
 		this.type = type;
 		this.name = name;
 		this.gender = gender;
 		this.legs = legs;
 		this.hands = hands;
 		this.says = says;
+		this.friends = friends;
 	}
 	toString(){
 		const props = ['type', 'name', 'gender', 'legs', 'hands', 'says'].map(prop => this[prop]);
@@ -74,6 +75,4 @@ man.friends = [dog, woman];
 catWoman.friends = [cat];
 
 // ======== OUTPUT ========
-[dog, cat, woman, man, catWoman].forEach(entity => {
-	print(entity.toString());
-});
+[dog, cat, woman, man, catWoman].forEach(entity => print(entity));
--- a/submissions/vv2529/a-tiny-js-world-oop/index.js
+++ b/submissions/vv2529/a-tiny-js-world-oop/index.js
@@ -7,17 +7,16 @@
 
 // ======== CLASS DEFINITIONS ========
 class Inhabitant {
-	constructor(type, name, gender, legs, hands, says, friends = []){
+	constructor(type, name, gender, legs, says, friends = []){
 		this.type = type;
 		this.name = name;
 		this.gender = gender;
 		this.legs = legs;
-		this.hands = hands;
 		this.says = says;
 		this.friends = friends;
 	}
 	toString(){
-		const props = ['type', 'name', 'gender', 'legs', 'hands', 'says'].map(prop => this[prop]);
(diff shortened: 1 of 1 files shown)


submissions/vv2529/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+dog.friends = [cat, woman, man];
+cat.friends = [woman, catWoman];
+woman.friends = [cat, man];
+man.friends = [dog, woman];
+catWoman.friends = [cat];
+
+// ======== OUTPUT ========
+[dog, cat, woman, man, catWoman].forEach(entity => {
+	print(entity.toString());
```
OleksiyRudenko: `toString` is a magic method, it gets called implicitly in a string context. So `print(entity)` would be sufficient.
Furthermore as we do not transform `entity` and also do not take any part of it but use it in its entirety `forEach(print)` is an effective equivalent of `forEach(entity => print(entity))`.

submissions/vv2529/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,79 @@
...
+class Human extends Inhabitant {
+	constructor(name, gender, says){
+		super('human', name, gender, 2, 2, says);
+	}
+}
+
+class CatHuman extends Human {
+	constructor(attachedTo, name, gender){
+		super(name, gender, attachedTo.says);
+		this.attachedTo = attachedTo;
```
OleksiyRudenko: es6 class syntax doesn't offer an elegant way for composition. Also inheritance and composition do not co-live well.

Anyway, borrowing anything from a peer class **instance** is not really OOP compliant.

If `Cat` had a method supplying the saying then you might borrow it right from `Cat`'s prototype chain.

Even more composition-like approach is to have class `Meower` both `Cat` and `CatHuman` could borrow from.

submissions/vv2529/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,79 @@
...
+		this.type = type;
+		this.name = name;
+		this.gender = gender;
+		this.legs = legs;
+		this.hands = hands;
+		this.says = says;
+	}
+	toString(){
+		const props = ['type', 'name', 'gender', 'legs', 'hands', 'says'].map(prop => this[prop]);
+		props.push(this.friends.map(item => item.name).join(', '));
```
OleksiyRudenko: class `Inhabitant` doesn't own `friends` property. So it is not OOP compliant to rely on that child classes or the outer world (even worse crime) would add any.
Please, fix.

submissions/vv2529/a-tiny-js-world-oop/index.js
vv2529: `forEach(print)` throws a `InvalidCharacterError`

vv2529:
Updated

submissions/vv2529/a-tiny-js-world-oop/index.js
```diff
@@ -0,0 +1,78 @@
...
+   */
+
+// ======== CLASS DEFINITIONS ========
+class Inhabitant {
+	constructor(type, name, gender, legs, hands, says, friends = []){
+		this.type = type;
+		this.name = name;
+		this.gender = gender;
+		this.legs = legs;
+		this.hands = hands;
```
OleksiyRudenko: Not every inhabitant has hands by natural design.
Imagine we add even-toed animals and animals like rhinos and kangaroos with their very special props. There is a risk the base class becomes cluttered very soon.

Please, make hands unique to humans.

Child class can overload base class' method where it would call a parent class method (`super.toString()`), supplement it with info specific to a given class and return aggregated result.

No problem if hands come after anything else. OOP compliance is more crucial in this task.

vv2529:
Updated. Keeping `hands` for animals really doesn't make sense.

PR #305, opened 2019-01-08
Memory pair

vv2529:
### Memory – Pair Game

[Demo](https://vv2529.github.io/memory-pair) 
[Code base](https://github.com/vv2529/memory-pair)

Please, review.


submissions/vv2529/memory-pair/script.js
```diff
@@ -0,0 +1,128 @@
...
+
+ function getId(id){return document.getElementById(id)}
+ function arrayShuffle(arr){return arr.sort(() => 0.5 - Math.random())};
+
+/*
+ * Events
+ */
+
+const board = getId('board'),
+modal = getId('modal'),
```
zonzujiro: You missed spaces here.

submissions/vv2529/memory-pair/script.js
```diff
@@ -0,0 +1,128 @@
+/*
+ * Helper functions
```
zonzujiro: This is a redundant comment. Comments must describe why something is going on. Not what. 

submissions/vv2529/memory-pair/script.js
```diff
@@ -0,0 +1,128 @@
...
+ */
+
+ function getId(id){return document.getElementById(id)}
```
zonzujiro: It is redundant. Pls, remove this function

submissions/vv2529/memory-pair/script.js
```diff
@@ -0,0 +1,128 @@
...
+ function arrayShuffle(arr){return arr.sort(() => 0.5 - Math.random())};
```
zonzujiro: If you want to create one-liner, use arrow function instead and drop `return` as a consequence.

submissions/vv2529/memory-pair/script.js
```diff
@@ -0,0 +1,128 @@
...
+flipped = [];
+let matched = 0,
+movesBest = movesCurrent = 0;
+
+function startGame(){
+	matched = movesCurrent = 0;
+
+	const frag = document.createDocumentFragment();
+	let arr = arrayShuffle(formArray());
+	arr.forEach(img => {
```
zonzujiro: You can use `append` method, which can add several elements in one operation. Without creating a `fragment`, etc.

submissions/vv2529/memory-pair/script.js
```diff
@@ -0,0 +1,128 @@
...
+	const flipper = makeElem('flipper');
+	const front = makeElem('front');
+	const back = makeElem('back');
+	const tileImg = makeElem('tile-img', 'img');
+	tileImg.src = `img/${img}.png`;
+	tileImg.alt = img;
+
+	tile.appendChild(flipper);
+	flipper.appendChild(front);
+	flipper.appendChild(back);
```
zonzujiro: You can remove this with one `append` call

submissions/vv2529/memory-pair/script.js
```diff
@@ -0,0 +1,128 @@
...
+	if((flipped.length >= 2) || (flipped[0] == elem) || elem.classList.contains('matched')) return;
+	flipped.push(elem);
+	elem.classList.add('flipped');
+
+	if(flipped.length == 2){
+		movesCurrentElem.textContent = ++movesCurrent;
+
+		if(getImg(flipped[0]) == getImg(flipped[1])){
+			flipped.forEach(elem => {
+				console.log(elem);
```
zonzujiro: You should remove all `console.log` from your code before creating PR

submissions/vv2529/memory-pair/script.js
```diff
@@ -0,0 +1,128 @@
...
+}
+function checkForWin(){
+	if(matched * 2 === boardWidth * boardHeight){
+		modal.classList.remove('hidden');
+
+		if(!movesBest || movesCurrent < movesBest){
+			movesBest = movesCurrent;
+			winMsg.textContent = 'New best!';
+		}
+		else winMsg.textContent = 'You win!';
```
zonzujiro: Something goes wrong with code-style here.

submissions/vv2529/memory-pair/script.js
vv2529: I use comments like this for navigation, not explanation, since they visually separate different chunks of code. For me it is convenient. Or is it a bad practice?

submissions/vv2529/memory-pair/script.js
vv2529: `getId` is much shorter to write than the full method, and it still clearly conveys what it does. Furthermore, in my PR #269 @AMashoshyna approved this approach. Could you explain why it should be removed, please?

submissions/vv2529/memory-pair/script.js
vv2529: Sorry, what spaces? Do you mean I should indent `modal`? I'm just asking to clarify what to do.

vv2529:
Updated!

vv2529:
Fixed! Prettier is a very useful tool! I don't like that it deletes multiple empty lines and inserts extra spaces in `if(...){` statements, but probably I should get used to it.
Thanks for your review!

PR #306, opened 2019-01-08
Add memory-pair-game files

webdevagent:
### memory-pair-game

[Demo](https://webdevagent.github.io/memory-cat-game) |
[Code base](https://github.com/webdevagent/frontend-2019-homeworks/blob/memory-pair-game/submissions/webdevagent/memory-pair-game/gameplay.js)

Please, review.

### DIFF
--- a/submissions/webdevagent/memory-pair-game/gameplay.js
+++ b/submissions/webdevagent/memory-pair-game/gameplay.js
@@ -12,12 +12,13 @@ const catGallery = [
   'cats/sweet-cat.jpeg',
 ];
 
-const gameCards = Array(12).fill(0);
-gameCards.forEach((num, i) => gameCards[i] = createBlock(catGallery[i % 6]));
+const gameCards = Array(catGallery.length * 2).fill(0);
+gameCards.forEach((num, i) => gameCards[i] = createBlock(catGallery[i % catGallery.length]));
 gameCards.forEach((num, i) => {
   num.querySelector('.flip-box-inner').dataset.order = i;
-  num.querySelector('.flip-box-front').dataset.order = i;
-  num.querySelector('.flip-box-front').dataset.compareNum = i % 6;
+  let frontCardData = num.querySelector('.flip-box-front').dataset;
+  frontCardData.order = i;
+  frontCardData.compareNum = i % catGallery.length;
 });
 galleryContainer.addEventListener('click', cardCompare)
 restartButton.addEventListener('click', startGame);
@@ -26,9 +27,10 @@ startGame();
 function startGame() {
   gameCards.sort(() => 0.5 - Math.random());
   gameCards.forEach(num => {
-    galleryContainer.appendChild(num)
-    num.querySelector('.flip-box-inner ').classList.remove('right', 'transform-card');
-    num.querySelector('.flip-box-inner ').classList.add('start');
+    galleryContainer.appendChild(num);
+    let innerCard = num.querySelector('.flip-box-inner');
+    innerCard.classList.remove('right', 'transform-card');
+    innerCard.classList.add('start');
     checkCard = null;
     canOpenCard = true;
   });
@@ -48,32 +50,29 @@ function createBlock(imgsrc) {
   let catImage = createCard('img');
   catImage.src = imgsrc;
   flipBoxBack.appendChild(catImage);
-  flipBoxInner.appendChild(flipBoxFront);
-  flipBoxInner.appendChild(flipBoxBack);
+  [flipBoxFront, flipBoxBack].forEach(num => flipBoxInner.appendChild(num));
   flipBox.appendChild(flipBoxInner);
   return flipBox;
 };
 
 function cardCompare({target}) {
-  let PlayCardClass = target.className;
+  let playCardClass = target.className;
   let targetInnerCard = galleryContainer.querySelector(`.flip-box-inner[data-order='${target.dataset.order}']`);
+  let canCompare = playCardClass != 'flip-box' && playCardClass != 'catGallery' && canOpenCard && playCardClass == 'flip-box-front';
   const sInterval = 1000;
   const lInterval = 2000;
 
-  function afterCompareAction(fcard, scard, className, time, action) {
+  function afterCompareAction(fcard, scard, className, time, compared) {
     setTimeout(() => {
-      if (action == 'add') {
-        fcard.classList.add(className);
-        scard.classList.add(className);
-      } else {
-        fcard.classList.toggle(className);
-        scard.classList.toggle(className);
+      fcard.classList.toggle(className);
+      scard.classList.toggle(className);
+      if (compared) {
         checkCard = null;
         canOpenCard = true;
       }
     }, time);
   }
-  if (PlayCardClass != 'flip-box' && PlayCardClass != 'catGallery' && canOpenCard && PlayCardClass == 'flip-box-front') {
+  if (canCompare) {
     if (checkCard == null) {
       targetInnerCard.classList.toggle('transform-card');
       checkCard = target;
@@ -82,13 +81,13 @@ function cardCompare({target}) {
     if (target.dataset.compareNum != checkCard.dataset.compareNum) {
       targetInnerCard.classList.toggle('transform-card');
       canOpenCard = false;
-      afterCompareAction(targetInnerCard, checkCardInnerCard, 'transform-card', 1000);
+      afterCompareAction(targetInnerCard, checkCardInnerCard, 'transform-card', sInterval, true);
     }
     if (target.dataset.compareNum == checkCard.dataset.compareNum && target != checkCard) {
       targetInnerCard.classList.toggle('transform-card');
       canOpenCard = false;
-      afterCompareAction(targetInnerCard, checkCardInnerCard, 'right', 1000, 'add');
-      afterCompareAction(targetInnerCard, checkCardInnerCard, 'start', 2000);
+      afterCompareAction(targetInnerCard, checkCardInnerCard, 'right', sInterval);
+      afterCompareAction(targetInnerCard, checkCardInnerCard, 'start', lInterval, true);
     }
   }
 };


zonzujiro:
image

I just clicked several times on every cards

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+//add addEventListeners to gameBlocks and restartButton
+gameCards.forEach(num => num.addEventListener('click', cardCompare));
+restartButton.addEventListener('click', startGame);
+
+//call function startGame() to  begin playing
+startGame();
+
+/*Initialize and define basic game methods
+Initialize and define function that randomize income Array*/
+function Shuffle(o) {
```
zonzujiro: This is a violation of naming convention - from uppercase should be named only constructors

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+  for (var j, x, i = o.length; i; j = parseInt(Math.random() * i), x = o[--i], o[i] = o[j], o[j] = x);
```
zonzujiro: Also, you have copy-pasted this code and it's unreadable right now ;)

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+
+//add addEventListeners to gameBlocks and restartButton
+gameCards.forEach(num => num.addEventListener('click', cardCompare));
+restartButton.addEventListener('click', startGame);
+
+//call function startGame() to  begin playing
+startGame();
+
+/*Initialize and define basic game methods
+Initialize and define function that randomize income Array*/
```
zonzujiro: This is a redundant comment. Comments must describe why something is going on. Not what. 

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+
+const gameCards = [];
+
+//fill ampty Array by game-blocks
+for (let i = 0; i < 12; i++) {
+  gameCards[i] = createBlock(catGallery[i % 6]);
+};
+
+//add addEventListeners to gameBlocks and restartButton
+gameCards.forEach(num => num.addEventListener('click', cardCompare));
```
zonzujiro: You should use event delegation where it is possible. Adding of listener to each element – is not very effective. 

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
+//Initialize and define basic game variables and objects
+const restartButton = document.querySelector('.restart');
+const galleryContainer = document.querySelector('.catGallery');
```
zonzujiro: Pls, use `kebab-case` or `snake_case` for classes. Not `camelCase`. It is for JS.

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+  'cats/looking-cat.jpg',
+  'cats/sleeping-cat.jpeg',
+  'cats/standing-cat.jpeg',
+  'cats/sweet-cat.jpeg',
+];
+
+const gameCards = [];
+
+//fill ampty Array by game-blocks
+for (let i = 0; i < 12; i++) {
```
zonzujiro: You should write in the declarative, not imperative way. 

Useful links: 

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+  for (var j, x, i = o.length; i; j = parseInt(Math.random() * i), x = o[--i], o[i] = o[j], o[j] = x);
+  return o;
+};
+
+//Game begining function defining
+function startGame() {
+  Shuffle(gameCards);
+  gameCards.forEach(num => {
+    galleryContainer.appendChild(num)
+    num.firstChild.classList.toggle('right', false);
```
zonzujiro: Why not `classList.remove`? O_o

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+};
+
+/*Initialize and define cardCompare function, that compare cards and run
+ internal method that based on the comparison result*/
+function cardCompare({target}) {
+  function afterCompareAction(className, time, action) {
+    setTimeout(() => {
+      if (action == 'add') {
+        checkCard.classList.add(className);
+        target.parentNode.classList.add(className);
```
zonzujiro: on't use constructions like `children[0]`, `parentNode`, `nextSibling`, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use `querySelector` or smth instead.

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+  };
+  let innerImg = target.parentNode.querySelector('img');
+  if (canOpenCard) {
+    if (checkCard == null) {
+      target.parentNode.classList.add('transformCard');
+      checkCard = target.parentNode;
+      checkCardImg = innerImg;
+    } else {
+      target.parentNode.classList.add('transformCard');
+      if (checkCardImg.src == innerImg.src && checkCardImg != innerImg) {
```
zonzujiro: Try to compare not by `src` of images, but add `data-*` to DOM elements and use it instead.

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,106 @@
...
+        canOpenCard = false;
+        afterCompareAction('right', 1000, 'add');
```
zonzujiro: Magic numbers are [bad](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)

webdevagent:
Thanks for reviewing @zonzujiro ! First time when I saw a bug that you find, I thought that it was a miracle)). But I think it was mistake associated with closures. So I fixed this bug(i hope this, tested so many times), and correct my code according to your remarks. I have been removed the redundant comments and magic numbers. Also, I have been replaced forLoops with forEach method, parentNode with querySelector. And started to compare gaming cards with data-*.  Thank for your remarks, and happy Winter holydays!)

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,94 @@
...
+  'cats/attention-cat.jpeg',
+  'cats/hungry-cat.jpeg',
+  'cats/looking-cat.jpg',
+  'cats/sleeping-cat.jpeg',
+  'cats/standing-cat.jpeg',
+  'cats/sweet-cat.jpeg',
+];
+
+const gameCards = Array(12).fill(0);
+gameCards.forEach((num, i) => gameCards[i] = createBlock(catGallery[i % 6]));
```
zonzujiro: `6` is what? Magic number! 😄 

Maybe `catGallery.length` will be better?

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,94 @@
...
+});
+galleryContainer.addEventListener('click', cardCompare)
+restartButton.addEventListener('click', startGame);
+startGame();
+
+function startGame() {
+  gameCards.sort(() => 0.5 - Math.random());
+  gameCards.forEach(num => {
+    galleryContainer.appendChild(num)
+    num.querySelector('.flip-box-inner ').classList.remove('right', 'transform-card');
```
zonzujiro: Looks like redundant space in selector. 

Also, no need to call `querySelector` twice. Call it once and save result in to variable

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,94 @@
...
+function createBlock(imgsrc) {
+  let flipBox = createCard('div', 'flip-box');
+  let flipBoxInner = createCard('div', 'flip-box-inner');
+  let flipBoxFront = createCard('div', 'flip-box-front');
+  let flipBoxBack = createCard('div', 'flip-box-back');
+  let catImage = createCard('img');
+  catImage.src = imgsrc;
+  flipBoxBack.appendChild(catImage);
+  flipBoxInner.appendChild(flipBoxFront);
+  flipBoxInner.appendChild(flipBoxBack);
```
zonzujiro: You can replace all `flipBoxInner.appendChild` with one call of `append`

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,94 @@
...
+  flipBox.appendChild(flipBoxInner);
+  return flipBox;
+};
+
+function cardCompare({target}) {
+  let PlayCardClass = target.className;
```
zonzujiro: You have violated name convention. From big letter you can call only names of the constructors

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,94 @@
...
+  let targetInnerCard = galleryContainer.querySelector(`.flip-box-inner[data-order='${target.dataset.order}']`);
+  const sInterval = 1000;
+  const lInterval = 2000;
+
+  function afterCompareAction(fcard, scard, className, time, action) {
+    setTimeout(() => {
+      if (action == 'add') {
```
zonzujiro: Can you replace this `if else` with simple `toggle`? 🤔

submissions/webdevagent/memory-pair-game/gameplay.js
```diff
@@ -0,0 +1,94 @@
...
+        scard.classList.add(className);
+      } else {
+        fcard.classList.toggle(className);
+        scard.classList.toggle(className);
+        checkCard = null;
+        canOpenCard = true;
+      }
+    }, time);
+  }
+  if (PlayCardClass != 'flip-box' && PlayCardClass != 'catGallery' && canOpenCard && PlayCardClass == 'flip-box-front') {
```
zonzujiro: Wow! Pls, can you save this check in some variable with a meaningful name? This will make code more readable.

webdevagent:
@zonzujiro, you catch me with one more magic number 😄. Thank's for again reviewing, I have been fixed all your remarks, and right now my code becomes cleaner and more readable.

PR #312, opened 2019-01-08
js-oop frogger-game

iamsokol:
### js-oop

[Demo](https://iamsokol.github.io/frogger-arcade-game/)  
[Code base](https://github.com/iamsokol/iamsokol.github.io/tree/master/frogger-arcade-game)

Please, review.

### DIFF
--- a/submissions/iamsokol/js-oop/app.js
+++ b/submissions/iamsokol/js-oop/app.js
@@ -1,58 +1,59 @@
 let allEnemies = [];
 
-const ENEMIES_POSITION = [50, 130, 220],
-    WIDTH_PLAYER = 65,
-    HEIGHT_PLAYER = 75,
+const ENEMIES_POSITION = [60, 140, 220],
+    WIDTH_PLAYER = 60,
+    HEIGHT_PLAYER = 70,
     WIDTH_BLOCK = 150,
     WIDTH_FIELD = 500,
+    RANDOM_INDEX = 125,
+    COEFFICIENT = 100,
+    TIME_DELAY = 1000,
     PLAYER_X = 200,
-    PLAYER_Y = 380;
+    PLAYER_Y = 380,
+    PLAYER_STEP_X = 102,
+    PLAYER_STEP_Y = 83;
 
-const Enemy = function(x, y, player) {
+let Character = function(x,y,sprite){
     this.x = x;
     this.y = y;
+    this.sprite = sprite;
+};
+
+Character.prototype.render = function() {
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@iamsokol 
This is not a code review but rather some impressions from demo.
When the hero stays on either of two topmost rows (marked with red rect on a screenshot below) and an enemy runs one row below the hero gets reset on enemy proximity.

image


The game also allows the hero to reach outside the playfield. Moreover the hero gets reset once outside the playfield. Not really clear if it is a win or not.

OleksiyRudenko:
@iamsokol I took a look at your code line 3, which says `const ENEMIES_POSITION = [50, 130, 220],`. 
The rows are all of equal height actually, while 130 - 50 === 80 and 220 - 130 = 90. This might be a source of somewhat unexpected behaviour.

submissions/iamsokol/js-oop/app.js
```diff
@@ -0,0 +1,101 @@
...
+
+Player.prototype.restartPlayer = function () {
+    player.x = PLAYER_X;
+    player.y = PLAYER_Y
+};
+
+Player.prototype.handleInput = function(pressedKey) {
+    switch (pressedKey) {
+        case 'left':
+            this.x -= this.motion + 50;
```
OleksiyRudenko: 50 (and 30 below, and also some other numbers across the code) is a mysterious number. Please, demystify defining it as a constant with a semantic name, just like you did for many other numbers.
We write code for other people, firsthand. Please, take care of possible future contributors to your code.

submissions/iamsokol/js-oop/app.js
```diff
@@ -0,0 +1,101 @@
...
+        this.x = -WIDTH_BLOCK;
+    }
+    this.checkCollisions();
+}
+
+Enemy.prototype.checkCollisions = function(dt) {
+    let position1 = player.x < this.x + WIDTH_PLAYER,
+        position2 = player.x + WIDTH_PLAYER > this.x,
+        position3 = player.y < this.y + HEIGHT_PLAYER,
+        position4 = player.y + HEIGHT_PLAYER > this.y;
```
OleksiyRudenko: Classes should be agnostic of the outer world. The only globals allowed to refer to from within the class are global constants and library functions.
Other entities should be known to some class when such class is explicitly informed about such entity.
Please, fix.

iamsokol:
@OleksiyRudenko, I fixed everything

PR #322, opened 2019-01-09
DOM-task implemented with array.prototype.find() method

thinker-tinker4:
### DOM

[Demo](https://thinker-tinker4.github.io/the_dom_task/) |
[Code base](https://github.com/thinker-tinker4/the_dom_task)

Please, review.


submissions/thinker-tinker4/the_DOM_task/app.js
```diff
@@ -0,0 +1,32 @@
+const ul = document.querySelector('.container');
+const h2 = document.querySelector('.techName');
+const p1 = document.querySelector('.description');
+const img = document.querySelector('.logo');
+const technology = (el) => {
```
AMashoshyna: Naming convention: functions names usually contain verbs. My suggestion is to call this function `setTechnology` to indicate that the function has side-effects (mutates data outside it's scope).

submissions/thinker-tinker4/the_DOM_task/app.js
```diff
@@ -0,0 +1,32 @@
...
+   document.h2 = el.name;
+   h2.content = el.name;
+   p1.textContent = el.paragraph;
+   img.src = el.image;
+};
+ul.addEventListener ('click', event => {
+   document.querySelector('.default').classList.remove('.default');
+   event.target.classList.add('default');
+   const currentParagraph = (el) => event.target.textContent == el.name;
```
AMashoshyna: Event handler will be executed on every click, so function `currentParagraph` will be created on every click too. This does not make sense. In is better to put this function outside event handler.

submissions/thinker-tinker4/the_DOM_task/app.js
```diff
@@ -0,0 +1,32 @@
...
+   technology(data.find(currentParagraph));
+ });
+let data = [
```
AMashoshyna: This data will not change, so please use `const` instead of `let` to indicate and to prevent this.

submissions/thinker-tinker4/the_DOM_task/index.html
```diff
@@ -0,0 +1,31 @@
...
+    </nav>
+    <div class="content">
+        <h2 class = "techName">React</h2>
+        <img class = "logo" src="img/react2.png">
+        <p class = "description">React makes it painless to create interactive UIs. Design simple views for each state in your application, and React will efficiently update and render just the right components when your data changes. Declarative views make your code more predictable and easier to debug.</p>
+    </div>
+</main>
+<script src="app.js"></script>
+</body>
+</html>
```
AMashoshyna: Please make sure to add newlines at the end of every file.

OleksiyRudenko:
@thinker-tinker4 Would you please rename the PR so it gives a better of what's inside in terms in tasks right on the PR listing page?
image

AMashoshyna:
Well done!

PR #325, opened 2019-01-10
add tiny js world by jsmuse

jsmuse:
### add tiny js world by jsmuse

[Demo](https://jsmuse.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/jsmuse/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/jsmuse/a-tiny-JS-world/index.js
+++ b/submissions/jsmuse/a-tiny-JS-world/index.js
@@ -1,0 +1,64 @@
+const catSay = 'meow';
+
+const dog = {
+    species: 'dog',
+    name: 'Toby',
+    gender: 'male',
+    legs: 4,
+    hands: 0,
+    saying: 'woof-woof!',
+};
+
+const cat = {
+    species: 'cat',
+    name: 'Markiz',
+    gender: 'male',
+    legs: 4,
+    hands: 0,
+    saying: catSay,
+    friends: ['Mouse', 'Tom'],
+};
+
+const woman = {
+    species: 'human',
+    name: 'Sarah',
+    gender: 'female',
+    legs: 2,
+    hands: 2,
+    saying: 'hey, sweetheart!',
+    friends: ['Rex', 'Tom', 'Jenn'],
+};
+
+const man = {
+    species: 'human',
+    name: 'Joe',
+    gender: 'male',
+    legs: 2,
+    hands: 2,
+    saying: 'hello there',
+};
+
+const catWoman = {
+    species: 'cat-woman',
+    name: 'Cat Woman',
+    gender: 'female',
+    legs: 2,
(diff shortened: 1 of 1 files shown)


submissions/jsmuse/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,64 @@
...
+    gender: 'female',
+    legs: 2,
+    hands: 2,
+    saying: catSay,
+};
+
+const listObj = (obj) => {
+    let str = '';
+
+    for (item in obj) {
```
OleksiyRudenko: The order of properties in an object is not guaranteed. Please, list `obj` properties explicitly and also employ `Array.join`.

submissions/jsmuse/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    return str;
+};
+
+print(listObj(dog));
+print(listObj(cat));
+print(listObj(woman));
+print(listObj(man));
+print(listObj(catWoman));
```
OleksiyRudenko: There is a pattern above.
Hint:
`[varA, varB].forEach(item => whateverFunction(whateEverAnotherFunction(item)));`

OleksiyRudenko: Please, also ensure your code files end with an empty line. It is important. You can tune your IDE to do this dirty job for you.
Not required to fix right now.

submissions/jsmuse/a-tiny-JS-world/index.js
jsmuse: This is made not for order. Zero is needed to hide keys with properties 0. I check if property not equal zero and show it.

jsmuse:
@OleksiyRudenko thanks for review. Please check again :)

submissions/jsmuse/a-tiny-JS-world/index.js
OleksiyRudenko: @jsmuse `for in`, `Object.keys` and other `Object` methods iterate or list object properties in some arbitrary order, there is no guarantee that e.g. `species` will come first, and `name` will follow `species`.
But we have to print properties in a very particular order. This means that we need to list properties explicitly.

To make the problem more straightforward imagine that you need to print all properties with an exception of `legs`.
How would you solve this task?
Please, suggest a fragment of code, i.e. `listObj` function.

OleksiyRudenko: @jsmuse A hint:
```
[
  varA,
  varB,
  obj.propA,
].join('; ');
```

jsmuse:
@OleksiyRudenko thank you for explain, I'am already fix this.

PR #333, opened 2019-01-10
Friends app

vv2529:
### Friends App

[Demo](https://vv2529.github.io/friends-app) 
[Code base](https://github.com/vv2529/friends-app)

Please, review.

### DIFF
--- a/submissions/vv2529/friends-app/index.html
+++ b/submissions/vv2529/friends-app/index.html
@@ -1,0 +1,61 @@
+<!doctype html>
+<html lang="en">
+<head>
+	<meta charset="utf-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<title>Friends app</title>
+	<link rel="stylesheet" href="css/style.css">
+</head>
+<body>
+	<header>Friends</header>
+	<div class="row page-content">
+		<aside class="sort">
+			<div class="sort-inner">
+				<fieldset>
+					<legend>Filter by...</legend>
+					<input class="field" type="text" id="input-search" placeholder="Name">
+					<div class="row">
+						<span>Gender:</span>
+						<select class="field" id="select-gender">
+							<option selected>all</option>
+							<option>male</option>
+							<option>female</option>
+						</select>
+					</div>
+				</fieldset>
+				<fieldset id="filter-set">
+					<legend>Sort by...</legend>
+					<div class="row">
+						<label>
+							<input type="radio" name="filter" value="name-asc">
+							Name A→Z
+						</label>
+						<label>
+							<input type="radio" name="filter" value="name-desc">
+							Name Z→A
+						</label>
+					</div>
+					<div class="row">
+						<label>
+							<input type="radio" name="filter" value="age-asc">
+							Age ↑
+						</label>
+						<label>
+							<input type="radio" name="filter" value="age-desc">
+							Age ↓
+						</label>
+					</div>
+				</fieldset>
+				<button class="field btn red" id="btn-reset">🗑 Reset</button>
+			</div>
+		</aside>
+		<main>
+			<div id="content"></div>
+			<button class="field btn" id="btn-load-more">Load more</button>
+		</main>
+	</div>
+	<footer>Made for Kottans frontend course — 2019</footer>
+	<script src="js/script.js"></script>
+	<script src="js/filter.js"></script>
+</body>
+</html>
--- a/submissions/vv2529/friends-app/js/filter.js
+++ b/submissions/vv2529/friends-app/js/filter.js
@@ -1,0 +1,60 @@
+let filters = {};
+
+function refreshCards() {
+	container.classList.add('invisible');
+	container.append(
+		...getUserCards(newUserData.length ? newUserData : userData)
+	);
+	if (newUserData.length) {
+		userData = newUserData.slice();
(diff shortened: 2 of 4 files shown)


submissions/vv2529/friends-app/index.html
```diff
@@ -0,0 +1,61 @@
...
+			</div>
+		</aside>
+		<main>
+			<div id="content"></div>
+			<button class="field btn" id="btn-load-more">Load more</button>
+		</main>
+	</div>
+	<footer>Made for Kottans frontend course — 2019</footer>
+	<script src="js/script.js"></script>
+	<script src="js/filter.js"></script>
```
zonzujiro: Can you replace this part with ES6 modules? 🤔

submissions/vv2529/friends-app/js/filter.js
```diff
@@ -0,0 +1,60 @@
...
+
+	shownUserData = userData.slice();
+	for (let key in filters) {
+		const fn = filters[key];
+		shownUserData = shownUserData.filter(fn);
+	}
+
+	userData.forEach(user => {
+		getUserCard(user).classList[
+			~shownUserData.indexOf(user) ? 'remove' : 'add'
```
zonzujiro: Better use `if` statement here. Also, you can use `includes` instead `indexOf`

submissions/vv2529/friends-app/js/filter.js
```diff
@@ -0,0 +1,60 @@
...
+		...getUserCards(newUserData.length ? newUserData : userData)
+	);
+	if (newUserData.length) {
+		userData = newUserData.slice();
+		newUserData.length = 0;
+	}
+
+	shownUserData = userData.slice();
+	for (let key in filters) {
+		const fn = filters[key];
```
zonzujiro: What `fn` is? IMO,  `shownUserData.filter(filters[key])` more understandable.

submissions/vv2529/friends-app/js/filter.js
```diff
@@ -0,0 +1,60 @@
...
+			~shownUserData.indexOf(user) ? 'remove' : 'add'
+		]('hidden');
+	});
+
+	container.classList.remove('invisible');
+}
+
+document.getElementById('input-search').addEventListener('input', function() {
+	if (this.value) {
+		filters.search = user => ~user.name.full.indexOf(this.value.toLowerCase());
```
zonzujiro: Same with `includes`

submissions/vv2529/friends-app/js/filter.js
```diff
@@ -0,0 +1,60 @@
...
+	} else delete filters.search;
+	refreshCards();
+});
+document.getElementById('select-gender').addEventListener('change', function() {
+	const val = this.selectedOptions[0].value;
```
zonzujiro: Better handle an event from the argument, not from `this`, because someday you will need to move your function to separate class or object. Or you will need to change the context of your call, then your function will be broken.

So relying on the argument is more obvious and safer.

submissions/vv2529/friends-app/js/filter.js
```diff
@@ -0,0 +1,60 @@
...
+	if (val != 'all') {
+		filters.gender = user => user.gender == val;
+	} else delete filters.gender;
+	refreshCards();
+});
+document
+	.getElementById('filter-set')
+	.addEventListener('click', function(event) {
+		const radio = event.target.closest('label').firstElementChild;
+		const fns = {
```
zonzujiro: What is `fns`?

submissions/vv2529/friends-app/js/script.js
```diff
@@ -0,0 +1,61 @@
+const container = document.getElementById('content');
+
+container.addEventListener('click', function(event) {
```
zonzujiro: `container.addEventListener('click', function({target}) {`

submissions/vv2529/friends-app/js/filter.js
```diff
@@ -0,0 +1,60 @@
...
+	const val = this.selectedOptions[0].value;
+	if (val != 'all') {
+		filters.gender = user => user.gender == val;
+	} else delete filters.gender;
+	refreshCards();
+});
+document
+	.getElementById('filter-set')
+	.addEventListener('click', function(event) {
+		const radio = event.target.closest('label').firstElementChild;
```
zonzujiro: Don't use constructions like `children[0]`, `firstElementChild `, `nextSibling`, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use `querySelector` or smth instead.

submissions/vv2529/friends-app/index.html
vv2529: I wanted to, but each module has its own scope, so variables from one cant be accessed from the other, which is a problem.

submissions/vv2529/friends-app/js/script.js
vv2529: I use `event.preventDefault()` here, not only `target`.

vv2529:
Fixed!
Sort by names and by ages? How is it possible? When one criteria is equal, then sort by the other, but it kinda works now like that.

zonzujiro:
My bad. Wrong PR.
Everything is ok ;)

PR #346, opened 2019-01-12
Building a Tiny JS World

vadymkotlyarenko:

### A Tiny JS World

[Demo](https://vadymkotlyarenko.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/vadymkotlyarenko/a-tiny-JS-world/blob/populate-world/index.js)

Please, review.

### DIFF
--- a/submissions/vadymkotlyarenko/a-tiny-js-world/index.js
+++ b/submissions/vadymkotlyarenko/a-tiny-js-world/index.js
@@ -1,4 +1,3 @@
-
 function capitalizeFirstLetter(str) {
     return str.charAt(0).toUpperCase() + str.slice(1);
 }
@@ -11,7 +10,7 @@ function Inhabitant(species, name, gender, legs, hands, saying, friends) {
     this.hands = hands;
     this.saying = saying;
     this.friends = friends;
-    this.printInformation = function() {
+    this.getTextForPrint = function() {
         var textToPrint = '';
         for (const key in this) {
             if (!this[key] || typeof(this[key]) === 'function') {
@@ -35,19 +34,17 @@ function Inhabitant(species, name, gender, legs, hands, saying, friends) {
             }
             textToPrint += capitalizeFirstLetter(key) + '=>' + this[key] + ';'
         }
-        print(textToPrint);
+        return textToPrint;
     };
 };
   
 const dog = new Inhabitant('dog', 'Buddy','male', 4, 0 , 'woof-woof');
-dog.printInformation();
+print(dog.getTextForPrint());
 const cat = new Inhabitant('cat', 'Garfield','male', 4, 0 , 'mew', dog);
-cat.printInformation();
+print(cat.getTextForPrint());
 const man = new Inhabitant('human', 'Jonathan','male', 2, 2 , 'Hello!',[dog,dog]);
-man.printInformation();
+print(man.getTextForPrint());
 const woman = new Inhabitant('human', 'Megan','female', 2, 2 , 'Hi!');
-woman.printInformation();
+print(woman.getTextForPrint());
 const catWomen = new Inhabitant('cat', 'Joe','female', 4, 0 , cat, dog);
-catWomen.printInformation();
-
-  
+print(catWomen.getTextForPrint());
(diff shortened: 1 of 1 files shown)


submissions/vadymkotlyarenko/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,53 @@
...
+                    textToPrint += capitalizeFirstLetter(key) + '=>' + this[key].name + ';'
+                }
+                continue;
+            }
+            if (this[key] === saying) {
+                textToPrint += capitalizeFirstLetter(key) + '=>' + (typeof(this[key]) === 'object' ? this[key].saying : this[key]) + ';'
+                continue;
+            }
+            textToPrint += capitalizeFirstLetter(key) + '=>' + this[key] + ';'
+        }
```
OleksiyRudenko: This block above looks a bit overcomplicated.
Why not just list properties explicitly apply modifying function as appropriate?
This will make code better readable.

Hint:
```
[
  varA,
  "text",
  `some text: ${varB}`,
  modify(varC),
  varD ? varD : modify(varD),
].join(delimiter);
```

Also take into consideration that the order of props in an object is not guaranteed, so `for ... in` and `Object` methods aren't the best options when a particular order of props traverse is required.

submissions/vadymkotlyarenko/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,53 @@
...
+        print(textToPrint);
```
OleksiyRudenko: `printInformation` does two important and very different jobs: (1) formatted string preparation and (2) media-specific output.
Just return a formatted string and apply `print` in the main job code fragment.

submissions/vadymkotlyarenko/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+function Inhabitant(species, name, gender, legs, hands, saying, friends) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.hands = hands;
+    this.saying = saying;
+    this.friends = friends;
+    this.printInformation = function() {
```
OleksiyRudenko: Not required to fix now.
Take into consideration that constructor functions like `Inhabitant` should have their methods defined using prototype chain, which is a topic of the next task theoretical part.

OleksiyRudenko:
@vadymkotlyarenko For every comment above:
- resolve conversation if you have implemented recommendation
- if any recommendation is not implemented, please, give a reason
- if anything is not clear, please, ask a question not resolving the conversation

The above workflow helps to keep communication open and clear.
Thanks.

submissions/vadymkotlyarenko/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,39 @@
...
+        `Friends => ${this.friends.length > 0 ? this.getFriendsName() : "No one"}`
+    ].join("; ")
+};
+Inhabitant.prototype.getFriendsName = function() {
+    let names = "";
+    this.friends.forEach(friend => {
+        names += friend.name + ", ";
+    });
+    return names;
+};
```
OleksiyRudenko: This function body can be shorter: `return this.friends.join(", ");`

submissions/vadymkotlyarenko/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,39 @@
...
+const dog = new Inhabitant("dog", "Buddy", "male", 4, 0, "woof-woof!");
+print(dog.getTextForPrint());
+const cat = new Inhabitant("cat", "Garfield", "male", 4, 0, "meu", dog);
+print(cat.getTextForPrint());
+const man = new Inhabitant("human", "Jonathan", "male", 2, 2, "Hello!", dog, cat);
+print(man.getTextForPrint());
+const woman = new Inhabitant("human", "Megan", "female", 2, 2, "Hi!");
+print(woman.getTextForPrint());
+const catWomen = new Inhabitant("cat", "Jane", "female", 4, 0, cat);
+print(catWomen.getTextForPrint());
```
OleksiyRudenko: If you first instantiate objects and then do print jobs, the printing part can be shorter and [DRYer](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself): `[dog, cat, man....].forEach(item => print(item.getTextForPrint());`

PR #350, opened 2019-01-12
Tiny World v2

CrazyPike:
### Tiny World with Objects

[Demo](https://github.com/CrazyPike/frontend-2019-homeworks/blob/master/submissions/CrazyPike/tiny-world-v2%239/index.html) |
[Code base](https://github.com/CrazyPike/frontend-2019-homeworks/tree/master/submissions/CrazyPike/tiny-world-v2%239)

Please, review.


OleksiyRudenko:
@CrazyPike Please, also remove all files except `index.js` as it is sufficient to conduct the code review.

OleksiyRudenko:
@CrazyPike "v2" in PR title is confusing as many submit remade tasks. Please rename the PR as per task name in the scope of rather frontend course than the repo the task uses for an assignment.

OleksiyRudenko:
@CrazyPike Do you need any support with changes implementation?

submissions/CrazyPike/tiny-world-v2#9/index.js
```diff
@@ -0,0 +1,89 @@
...
+    constructor(species,legs,name,gender,phrase){
+        this.species = species;
+        this.legs = legs;
+        this.name = name;
+        this.gender = gender;
+        this.phrase = phrase;
+    }
+
+    stringify() {
+        return [this.species, this.legs, this.hands, this.name, this.gender, this.phrase].filter(item => (item !== undefined )).join(';')}
```
OleksiyRudenko: The base class shouldn't know and manage props it doesn't own.
Developer may introduce other props specific to some classes. We do not need to clutter the base class with those. Instead we define and use specific props in child classes. That's the point of OOP.

So, remove hands from the base class. Overload (re-define) `stringify` method in `Human`.
It will call parent class method (`super.stringify()`), add info about hands and return aggregated result.

submissions/CrazyPike/tiny-world-v2#9/index.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+class Dog extends Being{
+    constructor(legs,name,gender,phrase) {
+        super('dog', legs,name,gender,phrase);
+
```
OleksiyRudenko: Please, remove unnecessary empty lines here and across the code to make code more compact and better readable.

submissions/CrazyPike/tiny-world-v2#9/index.js
```diff
@@ -0,0 +1,89 @@
...
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+
+print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
```
OleksiyRudenko: We do not need this plain output as it does nothing about OOP.

submissions/CrazyPike/tiny-world-v2#9/index.js
```diff
@@ -0,0 +1,89 @@
...
+beings.forEach(item => print(item.stringify()));
```
OleksiyRudenko: If you rename `stringify` into `toString` in all classes you will get a magic method.
`toString` is called implicitly whenever an object is used in a string context (and `print` uses its params in a string context).
So renaming would give us
`beings.forEach(item => print(item.toString()));`
As we can skip `toString` for it is called implictly we get
`beings.forEach(item => print(item));`
Now, as `print` matches callback specification and we do not modify `item` before passing it to `print` and also pass `item` entirely, not a property from it, we can just pass `print` as a callback function:
`beings.forEach(print);`

`.forEach` will take every element from the array and call the callback function (`print`) passing the element as an argument.

Look how laconic yet readable this code fragment becomes.

PR #354, opened 2019-01-12
Object oriented JS (frogger)

krash1408:
### Object oriented JS (frogger)

[Demo](https://krash1408.github.io/frogger-game/) |
[Code base](https://github.com/krash1408/frontend-2019-homeworks/tree/frogger/submissions/krash1408/frogger)

Please, review.


OleksiyRudenko:
@krash1408 
Please, check yourself against the list below and fix as appropriate.

   * [ ] there are core constants and derived constants
 * Requirements re **OOP**:
   * [ ] Separation of Concerns principle is followed
         (e.g. `update` method does only rendering and 
         doesn't **contain** any inline code to check e.g. collisions)

OleksiyRudenko:
@krash1408 The sum of your commits gives zero changes. Was it intended? Do you need any support with this PR?

Please, feel free asking anything here or in our dedicated chat.

krash1408:
I fixed the variable declaration. Please explain more about second section "Requirements re OOP:"

submissions/krash1408/frogger/js/app.js
```diff
@@ -0,0 +1,125 @@
...
+        this.speed = 100 + Math.floor(Math.random() * 222);
+    };
+
+    // Checks for collisions between the player and the enemies
+    if (player.x < this.x + 80 &&
+        player.x + 80 > this.x &&
+        player.y < this.y + 60 &&
+        60 + player.y > this.y) {
+        player.x = 202;
+        player.y = 405;
```
OleksiyRudenko: Classes shouldn't refer to any globals. The only exclusions are constants and library functions. `player` is a global variable.

How to resolve: pass player instance to every enemy explicitly.

submissions/krash1408/frogger/js/app.js
```diff
@@ -0,0 +1,125 @@
...
+    // Multiplies the speed by the dt parameter on the x axis
+    this.x += this.speed * dt;
+
+    // Once enemies are off the canvas, they reappear randomly with different speeds
+    if (this.x > 510) {
+        this.x = -50;
+        this.speed = 100 + Math.floor(Math.random() * 222);
+    };
+
+    // Checks for collisions between the player and the enemies
```
OleksiyRudenko: `update` does two important things:
- updates the character on the screen (note how this coincides with method name)
- checks for collision (note this is not reflected in method name)

Separation of Concerns principle requires to split important things into dedicated functions/methods. Single Responsibility principle from [SOLID](https://en.wikipedia.org/wiki/SOLID) says the same.

We could rename the method to `updateAndCheckCollisions` but this looks really weird and still breaks the Separation of Concerns principle.

submissions/krash1408/frogger/js/app.js
```diff
@@ -0,0 +1,125 @@
...
+    ctx.drawImage(Resources.get(this.player), this.x, this.y);
+};
+
+// Allows the user to use the arrow keys to jump from tile to tile
+Player.prototype.handleInput = function (keyPress) {
+
+    // Enables user on left arrow key to move left on the x axis by 102
+    // Also enables user not to go off the game tiles on the left side
+    if (keyPress == 'left' && this.x > 0) {
+        this.x -= 102;
```
OleksiyRudenko: `102` looks mysterious. Please, define a constant with an explanatory name and use it so contributors could easily understand what the value is about without reading much code.

Please, fix here and similar cases across the entire codebase.

Take a look at [**Requirements re Constants**](https://github.com/kottans/frontend-2019-homeworks/pull/354#issuecomment-453832868)

submissions/krash1408/frogger/js/app.js
```diff
@@ -0,0 +1,125 @@
...
+// Location of the 3 enemies on the y axis located on the stone road
+const enemyLocation = [63, 147, 230];
+
+
+// For each enemy located on the y axis from 0 on the x axis move at a speed of 200 
+// Until randomly regenerated in the enemy update function above
+enemyLocation.forEach(function (locationY) {
+    enemy = new Enemy(0, locationY, 200);
+    allEnemies.push(enemy);
+});
```
OleksiyRudenko: The fragment above converts some numbers into enemies with a particular locationY each.

When array gets converted it is a clear signal to employ `.map`.

When applied your code will look as short as
`let allEnemies = [63, 147, 230].map(locationY => new Enemy(....));`

Of course, you 63, 147, 230 are expected to be replaced with constants or a constant with some arithmetic to calculate these numbers.

submissions/krash1408/frogger/js/app.js
krash1408: This part is hard to understanding for me. I need an example.

submissions/krash1408/frogger/js/app.js
krash1408: I declarete a new prototype for Enemy, don't shure, that is right way to resolve this comment. Please, correct.

OleksiyRudenko: The way you solved the problem is correct. Collision is checked with a dedicated method, which gets called under proper circumstances (at every character location update).
Well done!

submissions/krash1408/frogger/js/app.js
OleksiyRudenko: The `player` within `Enemy` methods now is a global variable. This situation is not OOP compliant. The solution is to pass required object (global `player`) to an `Enemy` class instance at instantiation (`new Enemy(....)`) just like anything else, e.g. coordinates and speed is passed to it.
Does the above clarify?

krash1408: I guess, yes. I add a 'player' method to class Enemy and pass a player object when create an enemies. Is it right? Check, please.

PR #364, opened 2019-01-13
add "Friends App"

SergSenras:
### Friends App

[Demo](https://sergsenras.github.io/friends-app/) |
[Code base](https://github.com/SergSenras/SergSenras.github.io/tree/master/friends-app)

Please, review.

### DIFF
--- a/submissions/SergSenras/friends-app/script.js
+++ b/submissions/SergSenras/friends-app/script.js
@@ -1,0 +1,88 @@
+const CONTAINER = document.getElementById('container');
+const SIDEBAR = document.getElementById('sidebar');
+const SEARCH_FIELD = document.getElementById('searchField');
+const URL = 'https://randomuser.me/api/?results=40';
+
+let persons = [];
+let displayList = [];
+let searchStr = '';
+
+async function loadXMLDoc() {
+  let response = await fetch(URL)
+                  .then(data => data.json())
+                  .then(function(data){
+                    persons = data.results;
+                    displayList = persons;
+                    displayCards();
+                  })
+                  .catch(function(error){
+                    console.log('error');
+                  })
+}
+
+function getTemplate(el) {
+  return `<div class="card">
+            <div class="picture">
+              <img src="${el.picture.large}">
+            </div>
+            <h1>${el.name.last} ${el.name.first}</h1>
+            <div class="cardInfo">
+              <p><span class="text_title">Age:</span> ${el.dob.age}</p>
+              <p><span class="text_title">Gender:</span> ${el.gender}</p>
+              <p><span class="text_title">Phone:</span> ${el.phone}</p>
+            </div>
+          </div>`
+}
+
+function displayCards() {
+  CONTAINER.innerHTML = '';
+  let cardMarkup = '';
+  cardMarkup += displayList.reduce((accumulator, currentValue) => accumulator.concat(getTemplate(currentValue)), '');
+  CONTAINER.insertAdjacentHTML('afterbegin',cardMarkup);
+}
+
+document.addEventListener('DOMContentLoaded', function(){  
+  loadXMLDoc();
+
+  SIDEBAR.addEventListener('click', function(event){
+    let radio = event.target.closest('.rdBtn');
+    let btn = event.target.closest('button');
+    if(radio){
+      switch(radio.id){
+        case 'genderAll':
+          displayList = persons;
+          break;
+        case 'genderMale':
+          displayList = persons.filter( el => el.gender == 'male' );
+          break;
+        case 'genderFemale':
+          displayList = persons.filter( el => el.gender == 'female' );
+          break;
+        case 'nameAsc':
+          displayList = displayList.sort((b, a) => a.name.last > b.name.last ? -1 : 1);
+          break;
+        case 'nameDesc':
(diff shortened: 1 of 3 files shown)


submissions/SergSenras/friends-app/script.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+    else if(btn){
+      displayList = persons;
+    }
+    displayCards();
+  })
+
+  SEARCH_FIELD.addEventListener('keyup', function(event){
+    searchStr = SEARCH_FIELD.value;
+    displayList = persons.filter(el => el.name.last.indexOf(searchStr) != -1 || el.name.first.indexOf(searchStr) != -1)
```
AMashoshyna: Take a look at [String.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes)

submissions/SergSenras/friends-app/script.js
```diff
@@ -0,0 +1,88 @@
...
+          break;
+      }
+    }
+    else if(btn){
+      displayList = persons;
+    }
+    displayCards();
+  })
+
+  SEARCH_FIELD.addEventListener('keyup', function(event){
```
AMashoshyna: What if I paste the search phrase with Ctrl+V?

submissions/SergSenras/friends-app/script.js
```diff
@@ -0,0 +1,88 @@
+const CONTAINER = document.getElementById('container');
+const SIDEBAR = document.getElementById('sidebar');
+const SEARCH_FIELD = document.getElementById('searchField');
+const URL = 'https://randomuser.me/api/?results=40';
+
+let persons = [];
+let displayList = [];
+let searchStr = '';
```
AMashoshyna: Are you sure you need to make `searchStr` global variable?

submissions/SergSenras/friends-app/script.js
```diff
@@ -0,0 +1,88 @@
...
+
+async function loadXMLDoc() {
```
AMashoshyna: Could you please explain the choice of the name for this function? XML is a pretty specific name of a format.

submissions/SergSenras/friends-app/script.js
```diff
@@ -0,0 +1,88 @@
...
+  let response = await fetch(URL)
```
AMashoshyna: It is great that you have learned async functions, but in this particular case it does not seem to be a good use case for async/await. Your function doesn't return result of API query, so the code after this function call will run immediately. Usually async/await are most useful when you need to make several asynchronous operations sequentially and you don't want to use promise chains (they may look confusing indeed).

AMashoshyna:
Great job! Let's make some improvements.

User interface: if I enter empty string, search will be performed. Usually you should trim the input (remove whitespaces at the beginning and in the end of the string) and check for empty input.

Code formatting: please make code formatting (indentations and empty lines in particular) consistent. You can use your code editor's built-in formatting tools.

submissions/SergSenras/friends-app/script.js
SergSenras:
```suggestion
	SEARCH_FIELD.addEventListener('input', function (event) {
		let searchStr = '';
		searchStr = SEARCH_FIELD.value.trim();
		displayList = persons.filter(el => el.name.last.includes(searchStr));
```

submissions/SergSenras/friends-app/script.js
SergSenras:
```suggestion
  SEARCH_FIELD.addEventListener('input', function (event) {
```

submissions/SergSenras/friends-app/script.js
SergSenras: At first planned to use it in different places, that's why made global.
Fixed

submissions/SergSenras/friends-app/script.js
SergSenras: Well, that was the deadline )))
Fixed
```suggestion
function getPeopleData() {
```

submissions/SergSenras/friends-app/script.js
SergSenras: "Your function doesn't return result of API query" ©
why not?

Thanks for the explanations!
Fixed

SergSenras:
@AMashoshyna Made all requested chenges. Please review when you have a spare moment )

submissions/SergSenras/friends-app/script.js
AMashoshyna: The return value of `loadXMLDoc` is a Promise, but the value of this Promise is `undefined`. You put the result of API request into `response` variable, but you don't return it. It works for this particular case, but a more traditional approach for Promise chain is to return the result of the asynchronous action into next callback. Here is an article that helped me a lot: [We have a Problem with Promises](https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html)

AMashoshyna:
Well done!

submissions/SergSenras/friends-app/script.js
SergSenras: @AMashoshyna Thanks a lot!

PR #365, opened 2019-01-13
memory-game-task

sioniks:
### memory-game-task

[Demo](https://sioniks.github.io/memory-game/docs/index.html) |
[Code base](https://github.com/sioniks/memory-game)

Please, review.


submissions/sioniks/memory-game/index.js
```diff
@@ -0,0 +1,160 @@
...
+  }
+  // reset result
+  moves = 0;
+  counter.innerHTML = moves;
+  second = 0;
+  minute = 0;
+  hour = 0;
+  var timer = document.querySelector(".timer");
+  timer.innerHTML = "0 mins 0 secs";
+  clearInterval(interval);
```
IgorKurkov: move this code to separate function like resetResult - then you can delete your comment above - you will make more readable code, also use `let const`

submissions/sioniks/memory-game/index.js
```diff
@@ -0,0 +1,160 @@
...
+}
+
+var displayCard = function () {
+  this.classList.toggle("open");
+  this.classList.toggle("show");
```
IgorKurkov: why two open-action classes?

submissions/sioniks/memory-game/index.js
```diff
@@ -0,0 +1,160 @@
...
+  this.classList.toggle("disabled");
+};
+
+function cardOpen() {
```
IgorKurkov: also all the functions you can write on es6 notation, now you write all of them in global scope so you need only rewrite them to ES6

submissions/sioniks/memory-game/index.js
```diff
@@ -0,0 +1,160 @@
...
+    } else {
+      unmatched();
+    }
+  }
+}
+
+
+function matched() {
+  openedCards[0].classList.add("match", "disabled");
+  openedCards[1].classList.add("match", "disabled");
```
IgorKurkov: you can disable this action on js but its ok. In general idea with freezing cards trough js is better and cleaner, you can think about it

submissions/sioniks/memory-game/index.js
```diff
@@ -0,0 +1,160 @@
...
+function unmatched() {
+  openedCards[0].classList.add("unmatched");
+  openedCards[1].classList.add("unmatched");
+  disable();
+  setTimeout(function () {
+    openedCards[0].classList.remove("show", "open", "no-event", "unmatched");
+    openedCards[1].classList.remove("show", "open", "no-event", "unmatched");
+    enable();
+    openedCards = [];
+  }, 1100);
```
IgorKurkov: you can move this value to global config above (see first comment)

submissions/sioniks/memory-game/index.js
```diff
@@ -0,0 +1,160 @@
...
+}
+
+function disable() {
+  Array.prototype.filter.call(cards, function (card) {
```
IgorKurkov: you can use `Array#from`, also `[...cards]` do the same

submissions/sioniks/memory-game/index.js
```diff
@@ -0,0 +1,160 @@
...
+
+function moveCounter() {
+  moves++;
+  counter.innerHTML = moves;
+
+  if (moves == 1) {
+    second = 0;
+    minute = 0;
+    hour = 0;
+    startTimer();
```
IgorKurkov: not so obvious where do you get these values. see first comment

submissions/sioniks/memory-game/index.js
```diff
@@ -0,0 +1,160 @@
...
+
+function playAgain() {
+  popup.classList.remove("is-active");
+  startGame();
+}
+
+// loop to add event listeners to each card
+for (var i = 0; i < cards.length; i++) {
+  card = cards[i];
+  card.addEventListener("click", displayCard);
```
IgorKurkov: you need to keep in mind that make a listener for each element in a container is a bad idea, you need to use event delegation. but here you use 3 eventListeners on each element. please, rewrite this case to [event delegation](https://learn.javascript.ru/event-delegation), and think how you can reduce count of function-callbacks in one.

IgorKurkov:
@sioniks if you have any questions - feel free to ask :)

PR #369, opened 2019-01-13
firends-app

krash1408:
### firends-app
[Demo](https://krash1408.github.io/FriendsAPP/) |
[Code base](https://github.com/krash1408/frontend-2019-homeworks/tree/friends-app/submissions/krash1408/friends-app)

Please, review.


zonzujiro:
There are not only females here:

image

submissions/krash1408/friends-app/css/style.css
```diff
@@ -0,0 +1,277 @@
...
+    transition: background-color .5s;
+}
+.main-logo:hover,
+.main-logo:focus {
+    background-color: rgba(255, 255, 255, 0.925);
+    color: rgba(65, 166, 255, 0.856);
+    cursor: pointer;
+}
+
+/* поле сортировки */
```
zonzujiro: English comments only

submissions/krash1408/friends-app/src/index.js
```diff
@@ -0,0 +1,186 @@
...
+        usersCurrentArr.push(element);
+        usersSortedArr.push(element);
+      })
+    })
+    .then(function () {
+      createFiled( usersCurrentArr );
+    })
+    .catch(error => console.error(error));
+}
+//добавляет в массив к каждому обьекту ключ element со значением DOM елемента card
```
zonzujiro: This is a redundant comment. Comments must describe why something is going on. Not what. 

submissions/krash1408/friends-app/src/index.js
```diff
@@ -0,0 +1,186 @@
...
+function createCard ( value ) {
+  card = document.createElement('div');
```
zonzujiro: **Never** declare variables without an operator. This can cause leaking of variables into global scope. Also, it is unclear what are you doing - reuse existing variable or creating a new one.

submissions/krash1408/friends-app/src/index.js
```diff
@@ -0,0 +1,186 @@
...
+        usersSortedArr.push(element);
+      })
+    })
+    .then(function () {
+      createFiled( usersCurrentArr );
+    })
+    .catch(error => console.error(error));
+}
+//добавляет в массив к каждому обьекту ключ element со значением DOM елемента card
+function createCard ( value ) {
```
zonzujiro: This is a very large function. Split it, pls. Like `createFrontSide`, `createBackSide`, etc/

submissions/krash1408/friends-app/src/index.js
```diff
@@ -0,0 +1,186 @@
...
+  cardWrapper.appendChild(cardBack);
+
+  cardFront.appendChild(cardPhoto);
+  cardFront.appendChild(cardName);
+  cardFront.appendChild(cardAge);
+
+  cardBack.appendChild(cardBirthDay);
+  cardBack.appendChild(cardEmail);
+  cardBack.appendChild(cardID);
+  cardBack.appendChild(cardPhone);
```
zonzujiro: This calls can be replaced with one call of `append`

submissions/krash1408/friends-app/src/index.js
```diff
@@ -0,0 +1,186 @@
...
+
+  value.element = card;
+}
+
+function createFiled ( array ) {
+  array.forEach(value => fieldWrapper.appendChild(value.element))
+  field.appendChild(fieldWrapper);
+}
+// сортировщики
+let sortByAge = function( a, b ) {
```
zonzujiro: Such functions better to declare with `const`. In such a way you will avoid redeclaration of function.

submissions/krash1408/friends-app/src/index.js
```diff
@@ -0,0 +1,186 @@
...
+  cardBack.appendChild(cardBirthDay);
+  cardBack.appendChild(cardEmail);
+  cardBack.appendChild(cardID);
+  cardBack.appendChild(cardPhone);
+
+  value.element = card;
+}
+
+function createFiled ( array ) {
+  array.forEach(value => fieldWrapper.appendChild(value.element))
```
zonzujiro: This is also can be replaced with `append` and spread syntax.

submissions/krash1408/friends-app/src/index.js
```diff
@@ -0,0 +1,186 @@
...
+  let nameA = a.name.first.toLowerCase(), 
+      nameB = b.name.first.toLowerCase()
+  if (nameA < nameB) return -1
+  if (nameA > nameB) return 1
+  return 0
+}
+const sortByGenderMale = function  ( array ) {
+  let card = document.querySelectorAll('.card');
+  array.forEach((value, i) => {
+    if ( value.gender !== 'male') {card[i].style.display = 'none'}
```
zonzujiro: Don't use inline style changing. It is a bad approach for several reasons: 

 

1. First of all, a browser will apply such styling for each line separately/ Which means, that every such line of code will be a reason for running of calculations of a page and for drawing it,  so you can receive a performance problem. Read about reflow and repaint.  

2. This is an imperative way, you need to write declarative and describe what your code does, not how. This will make your code shorter and easier to maintain. 

3. Reuse of code. Saying, you will need to rotate some other stuff - you will add a similar line to another part of an application. Which is not right because of DRY. 

4. JS for logic, CSS - for styling. 

Also, it is better to filter your array and then redraw it, than just hide cards one-by-one.

submissions/krash1408/friends-app/src/index.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+})
+document.querySelector('#search').addEventListener('input', () => {
+  search( usersCurrentArr );
+})
+
+document.querySelectorAll('.card__front').forEach( val => {
+  val.addEventListener('click', () => {
+    console.log(target);
+    console.log(click)
```
zonzujiro: You should remove all your `console.log` prior PR.

krash1408:
I think, that's all now resolved, pls, review again.

PR #376, opened 2019-01-15
Friends app task

Roman-Halenko:
### Friends app
[Demo](https://roman-halenko.github.io/friends-app/) |
[Code base](https://github.com/Roman-Halenko/friends-app)

Please, review.

### DIFF
--- a/submissions/Roman-Halenko/friends-app/app.js
+++ b/submissions/Roman-Halenko/friends-app/app.js
@@ -32,9 +32,9 @@ function controller(srcArr) {
     renderHTML(list);
   }
 
-  function filterByGender(ev) {
-    makeActive(ev.target, config.filterBtns);
-    switch (ev.target.value) {
+  function filterByGender({target}) {
+    makeActive(target, config.filterBtns);
+    switch (target.value) {
       case 'male':
         list = srcArr.filter(e => e.gender === 'male');
         break;
--- a/submissions/Roman-Halenko/friends-app/app.js
+++ b/submissions/Roman-Halenko/friends-app/app.js
@@ -1,20 +1,24 @@
-const main = document.querySelector('main');
+const config = {};
+config.main = document.querySelector('main');
+config.sortBtn = document.getElementById('sort_btn');
+config.header = document.querySelector('header');
+config.searchField = document.getElementById('search');
+config.select = document.getElementById('input-sort');
+config.filterBtns = document.querySelectorAll('.btn_filter');
+config.filtersContainer = document.querySelector('.wrapper.filter_collection');
 
-const sortBtn = document.getElementById('sort_btn');
-
-const header = document.querySelector('header');
-
-const searchField = document.getElementById('search');
-
-const select = document.getElementById('input-sort');
-
-const filterBtns = document.querySelectorAll('.btn_filter');
-
-const filtersContainer = document.querySelector('.wrapper.filter_collection');
+function handleErrors(response) {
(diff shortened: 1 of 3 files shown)


submissions/Roman-Halenko/friends-app/app.js
```diff
@@ -0,0 +1,127 @@
...
+      <span class="label">Age: ${e.dob.age}</span>
+      <span class="location">
+        <i class="fa fa-home"></i> ${e.location.state}: ${e.location.city}
+      </span>
+    </div>
+    `);
+  })
+};
+
+// function compareNumbers(a, b) {
```
IgorKurkov: redundant comment

submissions/Roman-Halenko/friends-app/app.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+  });
+
+  renderHTML(list);
+};
+
+function renderHTML(arrOfObj) {
+  main.innerHTML = '';
+
+  arrOfObj.forEach(e => {
```
IgorKurkov: `arrOfObj` not so understandable, also `e`,  you can have a lot of data and different arrays of objects with diff data so better will be if you described more explanable your variables like userCards#forEach(card => {}) or whatever you want but it will be important in future

Roman-Halenko:
@IgorKurkov, thank, almost done :)
I'm not familiar yet with destructurisation. After some time on MDN, still havn't correct idea about implementatin in this case. I keep learn it. But, in fact, stuck a bit

submissions/Roman-Halenko/friends-app/app.js
```diff
@@ -1,47 +1,54 @@
...
   let inputStr = "";
 
   function searchByName(e) {
-    inputStr = searchField.value.toLowerCase();
+    inputStr = config.searchField.value.toLowerCase();
     list = srcArr.filter(e => e.name.last.includes(inputStr) || e.name.first.includes(inputStr));
     renderHTML(list);
   }
 
   function filterByGender(ev) {
```
IgorKurkov: about destructuring - you drop `event` as an argument - so if you don't use other props of `event` you can make a trick - `{ target }` in argument, and use this `target` variable only - it is the destructuring of objects in action

submissions/Roman-Halenko/friends-app/app.js
```diff
@@ -1,47 +1,54 @@
...
-
-const searchField = document.getElementById('search');
-
-const select = document.getElementById('input-sort');
-
-const filterBtns = document.querySelectorAll('.btn_filter');
-
-const filtersContainer = document.querySelector('.wrapper.filter_collection');
+const config = {};
+config.main = document.querySelector('main');
```
IgorKurkov: about this assign - I mean to make an obj and inside it make `prop: document.querySelector('some')` - this is a function so you will receive the desired element

submissions/Roman-Halenko/friends-app/app.js
```diff
@@ -56,72 +63,59 @@ function controller(srcArr) {
...
-      let nameA = a.name.first;
-      let nameB = b.name.first;
-      if (nameA > nameB) {
-        return 1;
-      }
-      if (nameA < nameB) {
-        return -1;
-      }
+      return a.name.first > b.name.first ? 1 : -1;
       return 0;
```
IgorKurkov: double `return` - will be uninvokable

PR #383, opened 2019-01-24
Move instructions for code review from kottans/frontend tasks

Oodmincheg:


Oodmincheg:
prizivayu @OleksiyRudenko!

README.md
```diff
@@ -23,9 +23,20 @@ This repo was created for students to submit their homeworks for review.
   - resolve any merge conflicts if there are any
   - push feature branch to your remote repository: `git push --set-upstream origin dom-api-task`
   - make pull-request from your repository to this repository via GitHub web-interface
-- wait for review from course mentors
-- if necessary, make changes, until your code will be approved and merged
-
+- post a link to your subtask PR in the
+     [FE Questionarium chat](https://t.me/joinchat/DmX0JAl-mh5W0jrWli8Ycw)
+     and ask mentors for a code review
+     * wait for review from course mentors
```
zonzujiro: Some tasks do not require review. If we want to collect everything in one file, we should mention this part here also.

README.md
```diff
@@ -23,9 +23,20 @@ This repo was created for students to submit their homeworks for review.
...
+     * if necessary, make changes, until your code will be approved and merged
+     * once the code review phase successfully finished
+     and you have an approval from one of the mentors
+     add to your `kottans-frontend/README.md`
+     a link to subtask code base and
+     a note saying `Reviewed and approved by @<mentor>`
+     (use mentor's github username)
+   * list your reflections on this subtask
+     (_what was new to you_, _what surprised you_, _what you intend to use in future_)
```
OleksiyRudenko: We ask for reflections in every task (non-PRable as well). If we keep this in `kottans/frontend` then no problem. If we want to mention this for PRable tasks here then we probably need to add that reflections should be placed in student's course repo (`kotans-frontend`).

What do you think?

PR #391, opened 2019-02-05
Google popup created

Roman-Halenko:
### Google-style Popup

[Demo](https://roman-halenko.github.io/html-css-popup) |
[Code base](https://github.com/Roman-Halenko/html-css-popup)

Please, review.


submissions/Roman-Halenko/html-css-popup/css/style.css
```diff
@@ -0,0 +1,192 @@
+body {
+  margin: 0;
+  height: 100vh;
+  background-color: #5e6969;
+  font-family: sans-serif;
+}
+
+header {
```
A-Ostrovnyy: Avoid styling by tag add custom styles on class.

submissions/Roman-Halenko/html-css-popup/css/style.css
```diff
@@ -0,0 +1,192 @@
...
+  height: 90%;
+  padding: 15px 25px;
+  overflow: scroll;
+}
+
+.menu_item {
+  text-align: center;
+}
+
+.menu_item:nth-child(1n+13) {
```
A-Ostrovnyy: Better use classname instead of `:nth-child(1n+13)`

Roman-Halenko:
@A-Ostrovnyy, hi! Thank for your review.
I've opened demo in 3 desktop's browsers and 1 mobile -  issue don't manifest itself.
How can I catch that?
> Also one item inside popup is missing [image](https://imgur.com/a/F9F6fBX)

A-Ostrovnyy:
In the Chrome on Windows all works fine... Some bug in Chrome on Mac OS don't worry about it. 
Fix another issues.
Also add `:hover` effect to list items.

submissions/Roman-Halenko/html-css-popup/css/style.css
```diff
@@ -5,7 +5,15 @@ body {
   font-family: sans-serif;
 }
 
-header {
+ul {
```
A-Ostrovnyy: You still set styles on tag.

Roman-Halenko: I left it consciously as it is a global rule that I would like to apply for all ul/li in my project. Should I resort to otherwise practice for this purpose?

A-Ostrovnyy: In this case, it will be enough. But better avoid styling by tag just use a class.
Rewrite your code or not it's up to you. In case you decide to rewrite your code let me know I will merge your homework or wait for your new code.

Roman-Halenko: Thank you for your attention :) I would like to left it as it is.

PR #420, opened 2019-02-11
Popup HTML/CSS

vv2529:
### Popup

[Demo](https://vv2529.github.io/kottans-practice/popup-no-js) 
[Code base](https://github.com/vv2529/kottans-practice/tree/master/popup-no-js)

Please, review.


submissions/vv2529/popup-no-js/source/stylus/extra/popup.styl
```diff
@@ -0,0 +1,73 @@
...
+	&::before
+		content ''
+		position absolute
+		display block
+		square(20px)
+		border 0 solid rgba(0,0,0,.1)
+		border-width 1px 0 0 1px
+		background-color #fff
+		transform translate(230px, -50%) rotate(45deg)
+	&-box
```
A-Ostrovnyy: Try to avoid writing a short selector. It's hard to find the right selector during the debugging code. Better write full selector: `.popup-box` instead of `.popup &-box`.
I mean when you copy full selector from browser dev tools and paste it into code editor you will find this selector but if you use the short name it not will be founded.
Also, short name save YOUR time but not a time of another developer who will work with your code after you.

vv2529:
Updated!

A-Ostrovnyy:
Good work!
Add outline/border on `:focus` state.
Also, button More, inside the popup, still without context(don't accessible from keyboard) and without outline/border too.
Please fix these issues.

vv2529:
I updated demo, all that should be fixed.

A-Ostrovnyy:
You updated demo but where is this commit in this repository?)

vv2529:
It is the commit I made yesterday

A-Ostrovnyy:
Nice work!

PR #421, opened 2019-02-12
popup

OlgaFrontend:
### Popup

[Demo](https://olgafrontend.github.io/kottans-frontend/popup_html_css/dist/) |  [Code](https://github.com/OlgaFrontend/kottans-frontend/tree/master/popup_html_css)

Please, review.


submissions/OlgaDubynka/popup/src/scss/_header.scss
```diff
@@ -0,0 +1,220 @@
...
+    width: 100%;
+    background-color: lightgray;
+  }
+
+  .list-more {
+    display: none;
+  }
+
+  .input-more {
+    visibility: hidden;
```
yevhenorlov: this way element can't receive focus. try hiding it via other methods so that "More" can be accessed from keyboard. How about adding `visually-hidden` class like so
```
.visually-hidden {
  position: absolute;
  clip: rect(0 0 0 0);
  width: 1px;
  height: 1px;
  margin: -1px;
}
```

and then - to show outline on the label when the checkbox is in focus:

```
.input-more:focus + .input-btn {
  outline: auto; // or any other outline you prefer
}

File: submissions/OlgaDubynka/popup/src/scss/_base.scss [hunk 2]
```diff
@@ -0,0 +1,14 @@
...
+  margin: 0;
+  padding: 0;
+  line-height: 1;
+  box-sizing: border-box;
+}
+
+a {
+  color: $grey;
+  cursor: pointer;
+}
```
A-Ostrovnyy: Add an empty line at the end of every file.

submissions/OlgaDubynka/popup/gulpfile.js
```diff
@@ -0,0 +1,85 @@
+var gulp      	= require('gulp');
+var scss        = require('gulp-sass');
+var imagemin    = require('gulp-imagemin');
+// var rename      = require('gulp-rename');
```
A-Ostrovnyy: Try not to commit the commented lines of code.

submissions/OlgaDubynka/popup/src/scss/_header.scss
OlgaFrontend: Thank you this advice :) I added this to my code and really it works better!

submissions/OlgaDubynka/popup/src/scss/_base.scss
OlgaFrontend: Yes, I agree with you about adding extra spaces at the end of each files, but I don't why sometimes one or two extra lines in my editor are not visible in github? Now I added more than one line, because if I add one I can it only in my editor, but not on Github.

submissions/OlgaDubynka/popup/gulpfile.js
OlgaFrontend: Oh yes it my faults! I promise do my best to prevent this bad habbit :)

A-Ostrovnyy: Use another extra branch for some experiments.)

submissions/OlgaDubynka/popup/src/scss/_base.scss
A-Ostrovnyy: Hiding an empty line at the end of the file is GitHub's behavior. One extra line is enough.

PR #431, opened 2019-02-16
html5-form ready for review

linkqwd:
[Demo](https://linkqwd.github.io/html5-order-form/) |
[Code base](https://github.com/linkqwd/html5-order-form/tree/master/src)

p.s. forgot about html\css validators. Will fix in a few days.


linkqwd:
Fixed HTML errors from w3c validator.
CSS has some errors because of features that are not fully supported at the current moment.

Results of form submit examples: https://webhook.site/#/9f86ffda-62e0-498b-97aa-fe5d72f5888b/fe20d352-a1e2-4d79-a637-035bae90729c/1

p.s. tried to use "normalize" instead of "reset". Won't make this mistake again.

submissions/linkqwd/html5-form/index.html
```diff
@@ -0,0 +1,453 @@
...
+                <div class="payment__relative-input">
+                  <label class="payment__label" for="card-code">Code</label>
+                  <input
+                    required
+                    class="payment__input payment__input_code"
+                    id="card-code"
+                    name="card-code"
+                    type="text"
+                    minlength="3"
+                    maxlength="3"
```
yevhenorlov: I believe this field should accept numbers only

linkqwd: added pattern for numbers

linkqwd:
Thanks for review!
I improved mobile version, but didn't understand this point

> * It would be better if the tick symbol on selected radio buttons appeared/disappeared together with the border

if there is need in further changes, please explain this once again.

submissions/linkqwd/html5-form/sass/_vars.sass
```diff
@@ -0,0 +1,117 @@
+@mixin media($width)
+  @media screen and (max-width: $width)
```
viktor-yakubiv: **Perhaps**, it's not very good idea to provide media queries for `screen` only.

Also, name of the mixin is too general; it's better to represent the functionality and call it like `max-screen-width`

submissions/linkqwd/html5-form/sass/_vars.sass
```diff
@@ -0,0 +1,117 @@
...
+$light-color: #9fa3c4
+$accent-color: #76a7e1
+$accent-dark-color: #416ea3
+$error-color: #e17777
+$gray-color: #f7f7f9
+$body-color: #eceef4
+$light-blue: #c9d0e1
+$orange: #edbe76
+
+.visually-hidden
```
viktor-yakubiv: If you call file `_vars.sass` that means file with variables only. Do not mix everything

The idea of the _variables_-files are in having general config in the single place and importing it in every explicitly or implicitly. If you import the same file explicitly in the every other file the compiler generates the same selectors as many times as you import it. Probably that's not what you want.

submissions/linkqwd/html5-form/sass/_vars.sass
```diff
@@ -0,0 +1,117 @@
...
+  position: absolute
+  clip: rect(0 0 0 0)
+  width: 1px
+  height: 1px
+  margin: -1px
+
+@mixin counter()
+	+media(800px)
```
viktor-yakubiv: Using tabs for indentation is good. However, tab size depends on environment and in Chrome it's 8 spaces, here at GitHub it's 8 spaces too, perhaps in your editor it's 2 space. When you mix tabs and spaces together it makes your code look inconsistently.

Most people prefer spaces because with spaces you can make custom spacing (not only increment by 2, 4 or 8 spaces) but it's up to you. Decide the convention for yourself.

submissions/linkqwd/html5-form/sass/styles.sass
```diff
@@ -0,0 +1,5 @@
+@import '_normalize.sass'
+@import '_vars.sass'
+@import '_range.sass'
+@import '_blocks.sass'
+
```
viktor-yakubiv: You see that you have a new line in the end but GitHub shows that you don't. Why?

Because you have a space or a few in the last line and the line is considered as not-empty. Set up trimming space in the end of line in your editor and you will fix this problem.

viktor-yakubiv:
I will provide functional review separately later. Good job! I really like it!

yevhenorlov:
> Thanks for review!
> I improved mobile version, but didn't understand this point
> 
> > * It would be better if the tick symbol on selected radio buttons appeared/disappeared together with the border
> 
> if there is need in further changes, please explain this once again.

there's an animation transition on the width of the border, but the tick symbol in the corner simply appears/disappears. this results in inconsistent animation (when you change focus via keyboard, tick appears first and then border fades in).

you can fix this by leaving the tick element in the DOM at all times and transitioning its opacity for on/off state (from completely transparent to completely visible).

this is not required, just a suggestion for better animation (and a matter of personal taste).

submissions/linkqwd/html5-form/sass/_vars.sass
linkqwd: Thanks! I will stick to 2 spaces indentation in future. For now I will keep it as it is. (it's a bit painfully to change indentation while using sass syntax)

linkqwd:
Thanks!
All done

PR #463, opened 2019-03-02
HTML5-Form

dnzln:
### HTML5-Form

[Demo](https://dnzln.github.io/html5-form/) |
[Code base](https://github.com/dnzln/html5-form)

Please, review.


submissions/dnzln/form/index.html
```diff
@@ -0,0 +1,477 @@
...
+                            Card number
+                    </label>
+                        <input type="text"
+                            class="card-number"
+                            name="card-number"
+                            id="card-number"
+                            title="Enter number of your card."
+                            minlength="16"
+                            maxlength="19"
+                            required>
```
yevhenorlov: 16 letters are valid, we should limit this to numbers only

submissions/dnzln/form/index.html
```diff
@@ -0,0 +1,477 @@
...
+                    </label>
+                        <input type="text"
+                            id="card-date"
+                            class="card-date"
+                            name="card-date"
+                            title="Enter expire date of your card: 01/24"
+                            placeholder="MM / YY"
+                            minlength="4"
+                            maxlength="5"
+                            required>
```
yevhenorlov: "aaaa" is a valid value. Also, title won't be shown because `minlength` and `maxlength` override it.
I would use `pattern` here (or `<input type="date">`, whichever one you prefer)

dnzln: I fixed validation, but don't get part about title - it showing up after hover, how length limitation can affect it?

dnzln:
Thanks for review! I thought I already made mobile version - is it so bad?)

submissions/dnzln/form/index.html
yevhenorlov: `title` also serves as text for validation popup when you use `pattern` validation

yevhenorlov:
> Thanks for review! I thought I already made mobile version - is it so bad?)
Screenshot_2019-03-15-11-07-36-460_com android chrome

Here's how it looks for me on android, not quite sure what's happening here :)

dnzln:
Hi! Sorry for long break. I added mobile version, please, take a look when you'll have time!

PR #471, opened 2019-03-15
WeatherApp

IrynaY:

### Weather-App

[Demo](https://irynay.github.io/weather-app/) |
[Code base](https://github.com/IrynaY/weather-app)

Please, review.


AMashoshyna:
Great job!
Here are some propositions for improvement.
1. CSS: when and item is added to Favourites, the middle part of the page moves left. It does not look like intended behavior =)
2. method  getCurrentWeather(city, units): I don't think you should mix getting the data and putting it into localStorage in one method. This code is hard to reuse if you need to fetch weather elsewhere in the application.
3. method getWeatherForecast(): I think last call of `then` is redundant - it return its parameter without doing anything
4. method subscribeForWeather(): it is often better to return Promise from the method, if it is created inside this method. What if you need to do something after Promise.all resolves?
5. This line of code is repeated multiple times:
```
WeatherDataService.subscribeForWeather(this.onServerResponse, this.state.name, this.state.units);
```
6. App component: in terms of React (and our framework follows this convention) props are received from parent component and should never be changed inside the component.

PR #473, opened 2019-03-18
The Weather App by feroxes

feroxes:
### The Weather App

[Demo](https://feroxes.github.io/the_weather_app/) |
[Code base](https://github.com/feroxes/the_weather_app)

Please, review.

*Because of google popup task, Google had blocked all my github pages. So you may see red screen with information about unsecure site

### DIFF
--- a/submissions/feroxes/the_weather_app/README.MD
+++ b/submissions/feroxes/the_weather_app/README.MD
@@ -1,0 +1,5 @@
+#The Weather Application
+
+<a href="https://feroxes.github.io/the_weather_app/">Demo</a>
+
+<a href="https://github.com/feroxes/the_weather_app">Source</a> 


yevhenorlov:
Nice job!

It feels like in some places you chose shorter syntax sacrificing readability instead. There's several improvements to that can be made, but in order to give you elaborate feedback I'm gonna need a PR of the code itself (to properly comment line-by-line).

However, here's a couple of notes straight away:

In slider file, there's a lot of ternary operators in places where if statements would be a better and more idiomatic solution:

```js
// don't:
n > slides.length ? slideIndex = 1 : null;
```

Besides syntax, they have different use cases: ternary operators are used for conditional logic that has some value to return (to assign to a variable, for example). For side effects, use if statements instead.

```js
// do:
if(n > slides.length) slideIndex = 1
```

Another example

```js
// don't:
n === (slides.length) ? next.style.display = 'none' : next.style.display = 'block';
```

This piece of logic is a valid usecase for a ternary operator, all we need to is change the side effect to an expression:

```js
// do:
next.style.display = n === (slides.length) ? 'none' :  'block';
```

Now we change the value depending on the condition and assign this value to `next.style.display`.

If you'd like to receive a detailed review, feel free to push a PR with all the code.

PR #475, opened 2019-03-26
WeatherApp ready for review

linkqwd:
[Demo](https://linkqwd.github.io/weather-app/)
[Code-base](https://github.com/linkqwd/weather-app/tree/master/src)

### known issues
- Timezone issue, currently daily forecast and sunrise\sunset-time shows correct data only for GMT+2 timezone.
Unfortunatley, used API doesn't have timezones info about requested region;
- Empty 'div' wrappers among layout which wraps Components (couldn't solve by my own);
- Tested and designed only in\for Chrome browser.

I can't fix those problems for now, due to lack of time, but hope to do it in future

### DIFF
--- a/submissions/linkqwd/WeatherApp/README.md
+++ b/submissions/linkqwd/WeatherApp/README.md
@@ -1,0 +1,10 @@
+# weather-app
+
+Demo: [link](https://linkqwd.github.io/weather-app/)
+Code-base: [link](https://github.com/linkqwd/weather-app/tree/master/src)
+
+### known issues
+- Timezone issue, currently daily forecast and sunrise\sunset-time shows correct data only for GMT+2 timezone.
+Unfortunatley, used API doesn't have timezones about requested region;
+- Empty 'div' wrappers among layout which wraps Components (couldn't solve by my own);
+- Tested and designed only in\for Chrome browser.


AMashoshyna:
Very good job! Here are come suggestions for improvements:

1. Async functions:
```javascript
  async getCurrentWeather(city) {
    const response = await fetch(
      `https://api.openweathermap.org/data/2.5/weather?q=${city}&units=metric&appid=3688267edbac491005d0dff07f9cbf22`
    );
    return await response.json()
  }
```
Except of some special cases (like try-catch block) combination of both `return` and `await` (last line of the method) is redundant. More on this [here](https://jakearchibald.com/2017/await-vs-return-vs-return-await/)

2. Single responsibility
Method `getTimeFromEpoch` should not be part of `Component` class. `Component` is part of the framework, which can potentially be reused in any application, so it should not contain any business logic. It can be part of application component or maybe `utils` module.

3. DOM abstraction should not be broken
`handleMenuOpenClose()` accesses the DOM directly, which is something that we want to avoid when using React-like library. This is more of jQuery approach. Please try to attach this callback in your `render` method. In this case you attach it to your VirtualDom, and the framework will handle the rest. you may need to add another wrapper element, but that's ok

Also, `body` is a misleading variable name, since there is `body` element on the page.

4. Calling `this.checkIfCityIsfav()` inside `render()` is not quite ok. `render()` should be pure function without side-effects. The best place for such actions is some component life-cycle method.

```javascript
    this.updateState({
      currentWeather: currentWeather,
      dailyForecast: dailyForecast,
      weeklyForecast: weeklyForecast
    });
```
5. This can be written shorter:
```javascript
    this.updateState({
      currentWeather,
      dailyForecast,
      weeklyForecast
    });
```

PR #488, opened 2019-04-14
Add weather-app

CuteShaun:
### weather-app

[Demo](https://cuteshaun.github.io/WeatherApp/) |
[Code base](https://github.com/CuteShaun/WeatherApp)

Please, review.


AMashoshyna:
Very well done! Let’s do some polishing.

1. Please avoid committing IDE-generated filtes (`.idea` folder)
2. Commented out code in production should be avoided unless absolutely necessary.
3. Abstract interaction with outer APIs:
- wrap interactions with localStorage in a service. Don’t forget to handle the errors
- avoid direct DOM manipulations – use virtualDOM to store and retrieve data instead of doing document.querySelector etc.
4. State should not be manipulated directly inside application component. Use dedicated framework methods for this.

AMashoshyna:
In WeatherDataService.js

```javascript
      const { longitude } = position.coords;
      const { latitude } = position.coords;
```

You can destructure multiple properties in one statement:

```javascript
	const { longitude, latitude } = position.coords;
```

AMashoshyna:
OpenWeatherMap URL is quite long and is used multiple times. Let’s put it in a constant, this will make code easier to read.

AMashoshyna:
```javascript
getWeatherByGeo(unit) {
//…
    return getGeoLocation()
      .then(getWeatherByPosition)
      .then(result => result);
}
```
Last line looks redundant

PR #495, opened 2019-04-25
Chat

IrynaY:
###  Chat again 😅

[Demo](https://secret-mountain-52099.herokuapp.com/) |
[Code base](https://github.com/IrynaY/frontend-2019-homeworks/tree/chat/submissions/IrynaY/chat)

Please, review.


AMashoshyna:
Wow, looks very nice! :fire: :fire: :fire:

I'm not sure if this is a bug or a feature, but if I'm alone in the chat, my messages will be forever pending. What do you think?

IrynaY:
@AMashoshyna Thanks for the review)
I think it's definitely a bug.
I tried to fix it, and for now, when a new user connect, he receives all unread messages, and the status of messages in the sender's chat also changes.

PR #498, opened 2019-04-27
add react app

AnnaGrynchuk:
### React Friends App

[Demo](https://annagrynchuk.github.io/React-Friends-App/) |
[Code base](https://github.com/AnnaGrynchuk/React-Friends-App)

Please, review.


submissions/AnnGrynchuk/react_app/src/SortButtons.js
```diff
@@ -0,0 +1,56 @@
...
+    @media screen and (max-width:400px) {
+    padding: 5px;
+    font-size:12px;
+    }
+`;
+
+const SortButtons = ({ handleSort }) => {
+    const onClick = (ev) => {
+        const id = ev.target.id;
+        handleSort({ id });
```
AMashoshyna: Let's give this key a more meaningful name than `id`. "sortType" maybe?

submissions/AnnGrynchuk/react_app/src/App.js
```diff
@@ -0,0 +1,128 @@
...
+    border-radius: 20px;
+    :hover {
+    background-color:#3AAAA1; 
+    }
+`;
+class App extends Component {
+  state = {
+    list: [],
+    name: '',
+    id: '',
```
AMashoshyna: `name` and `id` properties in state need better names IMO: name is actually a query string, because I can search for some characters and not the full name, right? As for id, as described below, it should be called something like "sortType"

submissions/AnnGrynchuk/react_app/src/App.js
```diff
@@ -0,0 +1,128 @@
...
+    result: [],
```
AMashoshyna: I don't see where you use `result` property. Looks like you calculate sorted arrays on the fly and don't save it in component's state (which is totally ok).

PR #510, opened 2019-05-10
React app

vladk96:
### React app

[Demo](https://vladk96.github.io/rick-and-morty-app/) |
[Code base](https://github.com/vladk96/rick-and-morty-app/tree/master/src)

Please, review.


vladk96:
@zonzujiro I don't know how to re-open the previous PR. So I create a new one with your remarks. In previous PR I resolved all conversations and left one comment.
Please, review.

submissions/vladk96/react-app/src/api/api.js
```diff
@@ -0,0 +1,41 @@
...
+
+const getCurrentPage = (pagePrevLink) => {
+  return pagePrevLink ? 1 : +(pagePrevLink + 1);
+}
+
+const createLinkForRequest = (objAttributes, url) => {
+  let searchParams = new URLSearchParams();
+
+  if (objAttributes) {
+    Object.keys(objAttributes).forEach(property => {
```
zonzujiro: URLSearchParams can accept object like an argument

submissions/vladk96/react-app/src/api/api.js
```diff
@@ -0,0 +1,41 @@
...
+
+const getData = async url => {
+  const response = await fetch(url);
+  const result = await response.json();
+  
+  return result;
+}
+
+const getCurrentPage = (pagePrevLink) => {
+  return pagePrevLink ? 1 : +(pagePrevLink + 1);
```
zonzujiro: More than max pages?

submissions/vladk96/react-app/src/api/api.js
```diff
@@ -0,0 +1,41 @@
+const URLS = {
+  characters: 'https://rickandmortyapi.com/api/character/?',
+  locations: 'https://rickandmortyapi.com/api/location',
+  episodes: 'https://rickandmortyapi.com/api/episode',
+}
+
+
+const getData = async url => {
+  const response = await fetch(url);
+  const result = await response.json();
```
zonzujiro: What if request will fail?

submissions/vladk96/react-app/src/api/api.js
```diff
@@ -0,0 +1,41 @@
...
+        searchParams.set(property, objAttributes[property]);
+      }
+    })
+  }
+
+  return url + searchParams.toString();
+}
+
+export const getCharacters = async (objAttributes, url = URLS.characters) => {
+  const createdUrl = await createLinkForRequest(objAttributes, url);
```
zonzujiro: It is not async. Pls, be attentive

submissions/vladk96/react-app/src/components/Characters/Characters.js
```diff
@@ -0,0 +1,136 @@
...
+      pages: list.info.pages,
+      currentPage: list.info.currentPage,
+      isLoaded: true
+    });
+  }
+
+  handlePaginationClick = async event => {
+    const clickedPageNumber = +event.target.textContent;
+
+    await this.setState({
```
zonzujiro: Why `await`?

submissions/vladk96/react-app/src/components/Characters/Characters.js
```diff
@@ -0,0 +1,136 @@
...
+    this.setState({
+      characters: list.results,
+      pages: list.info.pages,
+      currentPage: list.info.currentPage,
+      isLoaded: true
+    });
+  };
+
+  getObjectForRequest = ({ search, gender, species, currentPage }) => ({
+    name: search,
```
zonzujiro: I already asked, why not save this objects with params named as you need? Just to avoid renaming

submissions/vladk96/react-app/src/components/Characters/Characters.js
```diff
@@ -0,0 +1,136 @@
...
+  handleInputChange = event => {
+    const { name, value } = event.target;
+
+    this.setState({ [name]: value });
+  };
+
+  handleFormSubmit = async event => {
+    const reqAttributes = this.getObjectForRequest(this.state);
+
+    event.preventDefault();
```
zonzujiro: IMO, parent component must receive data from form, not event itself. And form should handle this logic.

submissions/vladk96/react-app/src/api/api.js
vladk96: You are right. This function works incorrectly. Moreover, this function is redundant because I change page numbers in Characters component and save page number in the state of it. So I delete it.

submissions/vladk96/react-app/src/components/Characters/Characters.js
vladk96: @zonzujiro I use `getObjectForRequest` function not to change property names, I use this func to create an object with attributes for request. Then `await getCharacters` function takes this object and create API link for request and return some data. I think this version is more clearly for reading.

My variant:
`const list = await getCharacters(this.getObjectForRequest(this.state));`

Another variant:
```
const list = await getCharacters({
  name: this.state.search,
  gender: this.state.gender,
  species: this.state.species,
  page: this.state.currentPage
  });
```
Or do you propose save this object in `state` and get it from `state`?

submissions/vladk96/react-app/src/components/Characters/Characters.js
vladk96: @zonzujiro If I got you correct, in `handleFormSubmit` function I must get attributes(searchValue, gender and species) not from `state`, must I get them from Form?

Should I use `new FormData(form)`?

zonzujiro: Just extract values from event into object and pass it into callback

vladk96: I must extract values from form in `Characters` and pass it to `Filters`. But how do I update `Characters` state with characters in `Filters` component?
Sorry I don't understand. Could you write more detail.

vladk96:
Please, review again. But the last conversation I haven't resolved. See my last comment in this conversation.

submissions/vladk96/react-app/src/components/Characters/Characters.js
zonzujiro: I am sorry, you decided to ignore the discussion?

submissions/vladk96/react-app/src/components/Characters/Characters.js
zonzujiro: I misunderstood code, my bad.

I thought you extracting values from event in  `handleFormSubmit`. I will resolve this one

submissions/vladk96/react-app/src/components/Characters/Characters.js
vladk96: Oh no. I decided that your option is better in the last comment, and I added it to my code. I apologize for closing the conversation.

submissions/vladk96/react-app/src/components/Characters/Characters.js
```diff
@@ -0,0 +1,140 @@
...
+
+  handlePaginationClick = event => {
+    const clickedPageNumber = +event.target.textContent;
+
+    this.setState(
+      {
+        isLoaded: false,
+        page: clickedPageNumber
+      },
+      async () => {
```
zonzujiro: The reason why you doing this in `setState` callback?

vladk96: `SetState` function works async.
So I used a callback in `setState` because `fetchCharacters` should be called after updating page number in `state` which the user clicked.

vladk96:
Hi. I decided to add `Loader` component as a container for content in `Characters` component. What do you think about it? Is it better solution?

submissions/vladk96/react-app/src/components/Characters/Characters.js
zonzujiro: What do you think, if you will place this logic into `componentDidUpdate` and `prevState.page !== this.state.page`?

In general, your approach is ok, but I prefer to avoid such code, because in a such way you rely on order state update outside lifecycle of component.

zonzujiro:
Loader is ok. 

I will merge it for now. It is complete. My last comment is up to you.

